PR #7, opened 2017-12-01
Irq rework master

samsor:
Rework of libmetal IRQ API:
 -Separate original and single metal_irq_register() into two independent functions metal_irq_register() and metal_irq_unregister().
 -Change Baremetal and FreeRTOS implementation to replace irq LUT and handler array with dynamically allocated containers.
 -Process optional device parameter for IRQ registration with Baremetal and FreeRTOS implementation.
 -Change Linux irq handling to check poll()'s revents POLLIN|POLLRDNORM instead of POLLIN|POLLERR
 -Improve irq unit tests to use new metal_irq_unregister() and add more coverage, also add Linux coverage.
 -Remove unneedeed irq_enable flag in generic and FreeRTOS irq lib
 -Change Linux implementation to replace irq handler array with dynamically allocated container.

### DIFF
--- a/lib/irq.h
+++ b/lib/irq.h
@@ -60,19 +60,33 @@ typedef int (*metal_irq_handler) (int irq, void *priv);
 struct metal_device;
 
 /**
- * @brief      Register interrupt or register interrupt handling of a
- *             specific interrupt.
+ * @brief      Register interrupt handler for driver ID/device.
  *
- *             If the interrupt handler parameter (irq_handler) is NULL,
- *	       deregister the interrupt handler.
+ * @param[in]  irq         interrupt id
+ * @param[in]  irq_handler interrupt handler
+ * @param[in]  dev         metal device this irq belongs to (can be NULL).
+ * @param[in]  drv_id      driver id is a unique interrupt handler identifier.
+ *                         It can also be used for driver data.
+ * @return     0 for success, non-zero on failure
+ */
+int metal_irq_register(int irq,
+		       metal_irq_handler irq_handler,
+		       struct metal_device *dev,
+		       void *drv_id);
+
+/**
+ * @brief      Unregister interrupt handler for driver ID and/or device.
  *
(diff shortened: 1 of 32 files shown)


lib/system/linux/device.c
```diff
@@ -305,7 +308,7 @@ static void metal_uio_dev_irq_ack(struct linux_bus *lbus,
 {
 	(void)lbus;
 	(void)irq;
-	unsigned int irq_info = 1;
+	int irq_info = 1;
 	unsigned int val;
```
wjliang: why change from int to "unsigned int", please revert this change as it doesn't look like relevant to this patch series

lib/irq.h
```diff
@@ -60,30 +60,44 @@ typedef int (*metal_irq_handler) (int irq, void *priv);
...
  * @param[in]  drv_id      driver id. It can be used for driver data.
  * @return     0 for success, non-zero on failure
  */
-int metal_irq_register(int irq,
-		       metal_irq_handler irq_handler,
-		       struct metal_device *dev,
-		       void *drv_id);
+int metal_irq_unregister(int irq,
+                         metal_irq_handler irq_handler,
+                         struct metal_device *dev,
```
wjliang: looks like white space difference.

lib/system/linux/device.c
samsor: the opposite, changed from unsigned int to int.

lib/system/linux/device.c
```diff
@@ -176,7 +176,7 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
 	struct dlist *dlist;
 	int result, i;
 	void *virt;
-	int irq_info;
+	unsigned int irq_info;
```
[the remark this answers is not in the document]
samsor: OK

lib/irq.h
samsor: OK

PR #10, opened 2017-12-14
Do memmap in io init

wjliang:
memory map the I/O region when creating the I/O region dynamically.
And some code cleanup


wjliang:
applied

PR #11, opened 2017-12-15
Fix sys irq save disable

samsor:
Fix issue that caused interrupts to not been disabled.
When cpsr.if was 0 and old value was 0, interrupt were not disabled.
This would happen if interrupt were initially enabled, after disabling, enabling and disabling again.
Changed sys_irq_save_disable() to read cpsr and compare interrupt bits to their disabled state, making sure interrupts gets disabled if different.


wjliang:
applied

PR #15, opened 2018-03-13
io: fix page_shift too big issue

wjliang:
When page_shift is larger than the number of bits of the offset
type, the offset shift operation will be undefined.
This fix detects if the page_shift is too big, if yes, the page
index will be "0".

### DIFF
--- a/lib/io.h
+++ b/lib/io.h
@@ -37,6 +37,7 @@
 #define __METAL_IO__H__
 
 #include <assert.h>
+#include <limits.h>
 #include <stdint.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


wjliang:
pull requests have been updated to address the comments:
* use sizeof(offset) instead of sizeof(metal_phys_addr_t)
* separate the changes into two commits

wjliang:
Thanks, I tested on generic and Linux before, i missed the RTOS. I have updated the patch and added the limits.h include

rfeneon:
Thanks!

PR #16, opened 2018-03-20
cache: Add a veneer to allow for inline implementations

galak:
To allow for smaller code or the ability to optimize away code that we
dont use add an veneer for the cache functions.  This way the system specific
code can decide how to implement the functions and do them completely inline
if desired.

### DIFF
--- a/lib/cache.h
+++ b/lib/cache.h
@@ -36,9 +36,12 @@
 #ifndef __METAL_CACHE__H__
 #define __METAL_CACHE__H__
 
+#include <metal/system/@PROJECT_SYSTEM@/cache.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
+
 
 /** \defgroup cache CACHE Interfaces
  *  @{ */
@@ -51,7 +54,10 @@ extern "C" {
  *                 If addr is NULL, and len is 0,
  *                 It will flush the whole data cache.
  */
-void metal_cache_flush(void *addr, unsigned int len);
+static inline void metal_cache_flush(void *addr, unsigned int len)
+{
+	return __metal_cache_flush(addr, len);
+}
 
 /**
  * @brief invalidate specified data cache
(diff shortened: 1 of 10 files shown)


galak:
RFC PR to see if this direction is desirable on what coding conventions we want to use for such veneers.

lib/system/zephyr/cache.h
```diff
@@ -29,21 +29,39 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline void __metal_cache_flush(void *addr, unsigned int len)
 {
-	metal_machine_cache_flush(addr, len);
+	metal_unused(addr);
+	metal_unused(len);
```
wjliang: Please add comments on the reason for not implementing the function.
Is it going to implement it in future?

lib/system/zephyr/cache.h
```diff
@@ -29,21 +29,39 @@
...
+	return;
 }
 
-void metal_cache_invalidate(void *addr, unsigned int len)
+static inline void __metal_cache_invalidate(void *addr, unsigned int len)
 {
-	metal_machine_cache_invalidate(addr, len);
+	metal_unused(addr);
+	metal_unused(len);
```
wjliang: Please add comments on the reason for not implementing the function.

lib/cache.h
```diff
@@ -36,10 +36,13 @@
 #ifndef __METAL_CACHE__H__
 #define __METAL_CACHE__H__
 
+#include <metal/system/@PROJECT_SYSTEM@/cache.h>
+
```
wjliang: This change will impact other OS implementation, the other OS needs cache.h too

galak: I will make the changes, but wanted to see if the general approach was ok first.

galak: For example should we use metal_machine_cache_flush or __metal_cache_flush

wjliang: This sounds a good approach and we can use this approach to improve some other APIs too.

lib/system/freertos/cache.c
```diff
@@ -38,12 +38,12 @@
...
 extern void metal_machine_cache_invalidate(void *addr, unsigned int len);
 
-void metal_cache_flush(void *addr, unsigned int len)
+void __metal_cache_flush(void *addr, unsigned int len)
 {
 	metal_machine_cache_flush(addr, len);
 }
 
-void metal_cache_invalidate(void *addr, unsigned int len)
+void __metal_cache_invalidate(void *addr, unsigned int len)
```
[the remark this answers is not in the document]
galak: Done

wjliang:
merged

PR #17, opened 2018-03-21
compiler: Fix restrict definition

galak:
The define of restrict was inside the c++ extern define block.  This
causes issues on older toolchains.

### DIFF
--- a/lib/compiler/gcc/compiler.h
+++ b/lib/compiler/gcc/compiler.h
@@ -38,9 +38,9 @@
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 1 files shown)


galak:
> should perhaps stick to the metal_ prefix to avoid colliding with application definitions...

That would be a follow up patch after this fix.  I'm fine with making that change as well.

wjliang:
as "restrict" is c99 standard, the issue to have this patch is because in there is older gcc compiler doesn't support by default and c++ . We can take this patch first and can have a follow up change.

PR #21, opened 2018-03-21
generic: Add a template files as starting point for a port

galak:
Add a template example for a generic platform to show what code needs to
get implemented by someone doing a port.  This also allows us to test
build the generic platform without any significant external
dependancies.

### DIFF
--- a/lib/system/generic/template/sys.h
+++ b/lib/system/generic/template/sys.h
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2018, Linaro Inc. and Contributors. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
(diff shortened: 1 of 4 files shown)


lib/system/generic/template/sys.h
```diff
@@ -0,0 +1,59 @@
...
+extern "C" {
+#endif
+
+#ifdef METAL_INTERNAL
+
+void sys_irq_enable(unsigned int vector);
+
+void sys_irq_disable(unsigned int vector);
+
+#endif /* METAL_INTERNAL */
```
wjliang: Good idea to have template.
We should also think about the other non-static functions implemented in the sys.c.
the benefit of sys_irq_enable|disable() being defined in this sys.h is the platform/OS dependent code can define if it is inline or not. This apply to all the others.
maybe change these two functions to be inline in the template?

galak: > Good idea to have template.
> We should also think about the other non-static functions implemented in the sys.c.
> the benefit of sys_irq_enable|disable() being defined in this sys.h is the platform/OS dependent code can define if it is inline or not. This apply to all the others.
> maybe change these two functions to be inline in the template?

I think we should do that as a follow up set of changes.  I was trying to get something in place  to allow CI to run for the generic platform.

PR #23, opened 2018-03-22
Ci generic

galak:
Update travis.yml to enable CI building off a generic target.


wjliang:
merged with added comment

PR #24, opened 2018-03-22
metal_phys_addr_t size depends on machine addr size

wjliang:
There can be 32bit machine with capability to access
64bit address. The metal_phys_addr_t size in this case
should be 64bit.

### DIFF
--- a/lib/sys.h
+++ b/lib/sys.h
@@ -50,7 +50,15 @@ extern "C" {
  *  @{ */
 
 /** Physical address type. */
+#ifdef ADDR_64BIT
+#warning "64bit address"
+typedef uint64_t metal_phys_addr_t;
+#elif ADDR_32BIT
+#warning "32bit address"
+typedef uint32_t metal_phys_addr_t;
+#else
 typedef unsigned long metal_phys_addr_t;
+#endif
 
 /** Interrupt request number. */
 typedef int metal_irq_t;


lib/sys.h
```diff
@@ -50,7 +50,15 @@ extern "C" {
  *  @{ */
 
 /** Physical address type. */
+#ifdef ADDR_64BIT
```
galak: Where is this define coming from?

wjliang: it will be from compilation flag.

wjliang:
the metal i/o region has physmap to store the physical address of the target. its type is metal_phys_addr_t, in such as system, we should use it to store 64bit address.

In shared memory case, it may not be required as it will not be very convenient for 32bit processor application to access the memory . However, in life cycle management case, the 32bit master host can load the remote to 64bit address space.

Or we can change the physmap in i/o region to be loaddr + high addr

starhwk:
Can this be made to 64bit as default?

The proposed approach doesn't seem quite straightforward. For example, how is application build supposed to figure out which type the libmetal on the system is built with? If such is needed, the exported header should rather be generated instead. Also, once this become configurable, it needs to make sure any variable with the type isn't shared directly across multiple platforms, otherwise there'll be unclear errors: ex, one platform giving 64bit type and the other reading as 32bit.

As far as the memory size concern, I don't see much references in the libmetal itself, other than region_phys, so the impact doesn't seem huge. I expect it's similar at the application side too. But I may be overlooking so please correct me if i'm wrong.

wjliang:
> Can this be made to 64bit as default?
> 
> The proposed approach doesn't seem quite straightforward. For example, how is application build supposed to figure out which type the libmetal on the system is built with? If such is needed, the exported header should rather be generated instead. Also, once this become configurable, it needs to make sure any variable with the type isn't shared directly across multiple platforms, otherwise there'll be unclear errors: ex, one platform giving 64bit type and the other reading as 32bit.
> 
> As far as the memory size concern, I don't see much references in the libmetal itself, other than region_phys, so the impact doesn't seem huge. I expect it's similar at the application side too. But I may be overlooking so please correct me if i'm wrong.

My concern on always using 64bit is on memory limited microcontrollers. It depends on how many memory regions an application have.
Agree it better not to introduce a new macro. Prefer to auto detect if it is 64bit or 32bit machine.

PR #27, opened 2018-03-29
Sleep veneer

galak:
To allow for smaller code or the ability to optimize away code that we
dont use add an veneer for the sleep function.  This way the system specific
code can decide how to implement the functions and do them completely inline
if desired.


wjliang:
merged

PR #28, opened 2018-03-29
compiler: introduce metal_weak macro

galak:
Introduce a macro to abstract how we define a weak function as this
tends to be compiler specific.


wjliang:
CI failed on Zephyr, compilation, is this issue there due to other compiler such as clang they have a different macro for it?

galak:
> CI failed on Zephyr, compilation, is this issue there due to other compiler such as clang they have a different macro for it?

No, its just that we needed a #include <metal/compiler.h>

wjliang:
merged

PR #30, opened 2018-03-30
Cfg irq

galak:
Changes to remove irq handler support from Zephyr port, as this doesn't make sense for Zephyr and provide a means via config.h for a user of libmetal to be able to test if the feature exists or not.


wjliang:
Should these configs be done in CMake file?

galak:
> Should these configs be done in CMake file instead of in config.h?

It seems like that config info would be static, so not sure what value doing it via CMake provides.

wjliang:
Just the configuration can be both dynamic and static (for those system which doesn't have that feature at all), and the CMakefile can be used to specify dynamic and static configurations. As we uses CMakefile to specify implementation (we specify which file to used) already, just seems like we can also uses CMakefile to specify both dynamic and static configuration options.

galak:
> Just the configuration can be both dynamic and static (for those system which doesn't have that feature at all), and the CMakefile can be used to specify dynamic and static configurations. As we uses CMakefile to specify implementation (we specify which file to used) already, just seems like we can also uses CMakefile to specify both dynamic and static configuration options.

I'm fine w/doing this with CMakefile, can you give me an example to see how to go about that.

wjliang:
> Just the configuration can be both dynamic and static (for those system which doesn't have that feature at all), and the CMakefile can be used to specify dynamic and static configurations. As we uses CMakefile to specify implementation (we specify which file to used) already, just seems like we can also uses CMakefile to specify both dynamic and static configuration options.
> 
> I'm fine w/doing this with CMakefile, can you give me an example to see how to go about that.

You can refer to camke/options.cmake

galak:
> Just the configuration can be both dynamic and static (for those system which doesn't have that feature at all), and the CMakefile can be used to specify dynamic and static configurations. As we uses CMakefile to specify implementation (we specify which file to used) already, just seems like we can also uses CMakefile to specify both dynamic and static configuration options.
> 
> I'm fine w/doing this with CMakefile, can you give me an example to see how to go about that.
> 
> You can refer to camke/options.cmake

Can you be more specific, is there a specific option that I can model this on?

wjliang:
> Just the configuration can be both dynamic and static (for those system which doesn't have that feature at all), and the CMakefile can be used to specify dynamic and static configurations. As we uses CMakefile to specify implementation (we specify which file to used) already, just seems like we can also uses CMakefile to specify both dynamic and static configuration options.
> 
>     I'm fine w/doing this with CMakefile, can you give me an example to see how to go about that.
> 
>     You can refer to camke/options.cmake
> 
> Can you be more specific, is there a specific option that I can model this on?
Here is what I am thinking about the change:
```
diff --git a/cmake/options.cmake b/cmake/options.cmake
index b7e5c44..3adb6df 100644
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -54,6 +54,9 @@ option (WITH_DOC "Build with documentation" ON)
 
 set (PROJECT_EC_FLAGS "-Wall -Werror -Wextra" CACHE STRING "")
 
+# Features Configuration
+option (WITH_IRQ_HANDLER "Build with IRQ handler" ON)
+
 check_include_files(xintc.h HAS_XINTC)
 if (HAS_XINTC STREQUAL "1")
   get_property(PROJECT_EC_FLAGS DIRECTORY PROPERTY PROJECT_EC_FLAGS)
diff --git a/lib/CMakeLists.txt b/lib/CMakeLists.txt
index 27c57ca..63af579 100644
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -60,6 +60,10 @@ if (WITH_DEFAULT_LOGGER)
   add_definitions (-DDEFAULT_LOGGER_ON)
 endif (WITH_DEFAULT_LOGGER)
 
+if (WITH_IRQ_HANDLER)
+       option(HAS_METAL_IRQ_HANDLER "" ON)
+endif (WITH_IRQ_HANDLER)
+
 if (WITH_ZEPHYR)
   zephyr_library_named(metal)
   add_dependencies(metal offsets_h)
diff --git a/lib/config.h b/lib/config.h
index cab1874..445a02d 100644
--- a/lib/config.h
+++ b/lib/config.h
@@ -67,6 +67,8 @@ extern "C" {
 #cmakedefine HAVE_STDATOMIC_H
 #cmakedefine HAVE_FUTEX_H
 
+#cmakedefine HAS_METAL_IRQ_HANDLER
+
 #ifdef __cplusplus
 }
 #endif
diff --git a/lib/system/zephyr/CMakeLists.txt b/lib/system/zephyr/CMakeLists.txt
index 981e790..8e1a424 100644
--- a/lib/system/zephyr/CMakeLists.txt
+++ b/lib/system/zephyr/CMakeLists.txt
@@ -15,6 +15,8 @@ collect (PROJECT_LIB_SOURCES shmem.c)
 collect (PROJECT_LIB_SOURCES sleep.c)
 collect (PROJECT_LIB_SOURCES time.c)
 
+SET (WITH_IRQ_HANDLER OFF PARENT_SCOPE)
+
 if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_MACHINE})
 endif (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_MACHINE})
```

galak:
> Here is what I am thinking about the change:

Thanks. I've integrated that change and updated the PR.

wjliang:
Just another thing to consider, even zephyr doesn't allow to register, enable/disable a particular interrupt at run time, however, it should have disable/restore global interrupt at run time. Instead of having this configuration option, shall we keep the top level wrapper, for those function which doesn't support in Zephyr, have assert in the function?

galak:
> Just another thing to consider, even zephyr doesn't allow to register, enable/disable a particular interrupt at run time, however, it should have disable/restore global interrupt at run time. Instead of having this configuration option, shall we keep the top level wrapper, for those function which doesn't support in Zephyr, have assert in the function?

I'm not sure I see the utility in exposing what ends up being zephyr driver model specific details if we can't abstract them fully via libmetal.

galak:
> However, zephyr should allow disable/restore interrupt globally at runtime?

Yes, and I left that code alone. 

lib/system/zephyr/irq.c Still has code for:

```
unsigned int metal_irq_save_disable(void)
void metal_irq_restore_enable(unsigned int flags)
void metal_irq_enable(unsigned int vector)
void metal_irq_disable(unsigned int vector)
```

galak:
> It just like some of the IRQ APIs not available for Zephyr, it makes the config option confusing. why not put empty implementation to metal_irq_register() and metal_irq_register() or define them as empty:
> 
> #define _metal_irq_register(...)
> ?

I think having metal_irq_register() do nothing on Zephyr is misleading to the user of libmetal.  I think its better that the API doesn't exist as this creates an explicit build issue for the user.  If someone takes some code that was calling metal_irq_register() and than tried using it on Zephyr I'd rather them get a build error and know they can't just use the code directly, vs a silent issue when the build was succeed but some behavior of the code they tried to use didn't work.

galak:
> Just another view of the metal_irq_register(), it provides the capability to register handler to this IRQ event at run time. That is can allow user to register metal_irq_isr from the start of the application and this isr will mark this interrupt is raised if it is called . And then at run time, to the the application, user can choose to register and deregister the handling to this event. We can provide this capability to user from the metal irq abstraction.

First, I really don't see the point in providing the abstraction in libmetal to begin with for space constraint systems.

As for implementing register/deregister on Zephyr, we would still have a piece of code missing that the user would have to provide.  It seems like the whole point of the abstraction in libmetal is so that a user doesn't have to provide additional bits of code, they just program to the abstraction.  If one some systems they need extra code and on others they don't what value is that providing?

Regardless of support for metal_irq_{de,}register, I think the feature should be able to be compiled out as its adds code on space constraint systems that might not be needed if one isn't using the feature.

wjliang:
> First, I really don't see the point in providing the abstraction in libmetal to begin with for space constraint systems.

The libmetal is the porting layer for the library across different systems, otherwise the solution will be limited to one type of system. We just moved the porting layer out of OpenAMP code base and make the OpenAMP library to focus on the specification/protocol implementation. 

> 
> As for implementing register/deregister on Zephyr, we would still have a piece of code missing that the user would have to provide. It seems like the whole point of the abstraction in libmetal is so that a user doesn't have to provide additional bits of code, they just program to the abstraction. If one some systems they need extra code and on others they don't what value is that providing?

I understand that some cases, there is no need to register the interrupt handler at run time, but this API just provide user an option to in case they want to use it.
The OpenAMP code before we moved the porting layer to libmetal also had the IRQ register abstraction in its porting layer as it allows OS independent platform drivers code to register OpenAMP interrupt handler during driver initialization.

> 
> Regardless of support for metal_irq_{de,}register, I think the feature should be able to be compiled out as its adds code on space constraint systems that might not be needed if one isn't using the feature.

If the application doesn't use the IRQ registration API, the code will not be linked to the application. If it is not implemented in zephyr support in libmetal while user tries to use it, it will give a link error. 
About the code space, I think it is more related to the init code. If Zephyr doesn't need to have the IRQ related initialization, the Zephyr IRQ init can be removed.

galak:
> The libmetal is the porting layer for the library across different systems, otherwise the solution will be limited to one type of system. We just moved the porting layer out of OpenAMP code base and make the OpenAMP library to focus on the specification/protocol implementation.

I understand, I just felt that IRQ registration/deregistration APIs feels beyond the line of abstraction.  I think integration of interrupt handlers tends to be viewed as system specific so attempting to abstract it becomes difficult, or re-implementing a lot of infrastructure outside of what a given OS/RTOS provides.


> I understand that some cases, there is no need to register the interrupt handler at run time, but this API just provide user an option to in case they want to use it.
> The OpenAMP code before we moved the porting layer to libmetal also had the IRQ register abstraction in its porting layer as it allows OS independent platform drivers code to register OpenAMP interrupt handler during driver initialization.

I understand, but my point is that the abstraction is not complete on Zephyr and never can be in just libmetal.  So what exact expectation are we setting for the user when they call metal_irq_register()?  Maybe my understanding of what the API is meant to accomplish is incorrect.

> If the application doesn't use the IRQ registration API, the code will not be linked to the application. If it is not implemented in zephyr support in libmetal while user tries to use it, it will give a link error.
> About the code space, I think it is more related to the init code. If Zephyr doesn't need to have the IRQ related initialization, the Zephyr IRQ init can be removed.

What would we use to make this choice?  a config option?

wjliang:
> 
> I understand, but my point is that the abstraction is not complete on Zephyr and never can be in just libmetal. So what exact expectation are we setting for the user when they call metal_irq_register()? Maybe my understanding of what the API is meant to accomplish is incorrect.

Just the libmetal provides a common abstraction, for some OS, if the abstraction is not complete, we can either provide implementation in libmetal or leave the implementation in libmetal not implemented.
 
> 
> What would we use to make this choice? a config option?
just a config option doesn't sounds very good as very API in the common abstraction layer can be potentially not have implementation in some system. And the solution depends on the purpose, as the patch purpose is for user to be notified that these implementation is not available  for Zephyr at build time. I think can just leave it as not implemented. In this case, if the user app calls it, it will give a linking error.

galak:
> Just the libmetal provides a common abstraction, for some OS, if the abstraction is not complete, we can either provide implementation in libmetal or leave the implementation in libmetal not implemented.

So what is the exact expectation of what metal_register_irq() is suppose to do for the user.  I ask because it seems like the only platform we can provide a full implementation on is Linux.  Which makes me question the ability the abstract this and thus the utility of trying to do that in libmetal.



> just a config option doesn't sounds very good as very API in the common abstraction layer can be potentially not have implementation in some system. And the solution depends on the purpose, as the patch purpose is for user to be notified that these implementation is not available for Zephyr at build time. I think can just leave it as not implemented. In this case, if the user app calls it, it will give a linking error.

Thus, we have both .config in which one case test against HAS_METAL_IRQ_HANDLER, and not having metal_irq_register exist at all (thus causing link error).

galak:
Superceded by PR #42

PR #31, opened 2018-03-30
lib: zephyr: noop metal_generic_dev_sys_open

galak:
Since Zephyr is running bare metal there isn't any mapping of io regions
that needs to be done, so we can have metal_generic_dev_sys_open do
thing to reduce code size


wjliang:
merged

PR #36, opened 2018-04-05
Zephyr logging

skordal:
Adds a default log handler for Zephyr applications, which is used unless another log handler is passed through the init params.


galak:
I didn't get a chance to see this, but wondering if instead of printk, we should use Zephyr's syslog functionality in include/logging/sys_log.h

skordal:
I think that should be possible, and I can probably make an update for it later. What I added in this PR is basically the same as what was already in the test application.

galak:
Sounds good, I opened issue #38 to track it.

PR #37, opened 2018-04-05
Metal assertions

skordal:
This PR proposes to add a `metal_assert()` macro to libmetal. This allows applications to avoid the dependency on the implementation of `assert()` from the standard library, which can pull in functions from [stdio](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=newlib/libc/stdlib/assert.c) or other places depending on the library.

This PR links `metal_assert()` to the `__ASSERT()` macro when used for Zephyr or the standard libc `assert()` otherwise.

This will also be useful when cleaning up the assert usage in OpenAMP, where several different macros (with different implementations) are used.

### DIFF
--- a/lib/assert.h
+++ b/lib/assert.h
@@ -11,7 +11,7 @@
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
  *
- * 3. Neither the name of Linaro nor the names of its contributors may be used
+ * 3. Neither the name of Xilinx nor the names of its contributors may be used
  *    to endorse or promote products derived from this software without
  *    specific prior written permission.
  *
@@ -29,36 +29,20 @@
  */
 
 /*
- * @file	zephyr/log.c
- * @brief	Zephyr libmetal log handler.
+ * @file	assert.h
+ * @brief	Assertion support.
  */
 
-#include <stdarg.h>
-#include <metal/log.h>
-#include <zephyr.h>
+#ifndef __METAL_ASSERT__H__
+#define __METAL_ASSERT__H__
 
-static const char *level_strs[] = {
-	"metal: emergency: ",
-	"metal: alert:     ",
-	"metal: critical:  ",
-	"metal: error:     ",
-	"metal: warning:   ",
-	"metal: notice:    ",
-	"metal: info:      ",
-	"metal: debug:     ",
-};
+#include <metal/system/@PROJECT_SYSTEM@/assert.h>
 
-void metal_zephyr_log_handler(enum metal_log_level level,
-			      const char *format, ...)
-{
-	va_list args;
+/**
+ * @brief Assertion macro.
+ * @param cond Condition to test.
+ */
+#define metal_assert(cond) metal_sys_assert(cond)
 
-	if (level <= METAL_LOG_EMERGENCY || level > METAL_LOG_DEBUG)
-		level = METAL_LOG_EMERGENCY;
-	printk("%s", level_strs[level]);
-
-	va_start(args, format);
-	vprintk(format, args);
-	va_end(args);
-}
+#endif /* __METAL_ASSERT_H__ */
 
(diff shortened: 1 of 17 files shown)


lib/assert.h
```diff
@@ -0,0 +1,49 @@
...
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Xilinx nor the names of its contributors may be used
```
wjliang: The should the license text be "Xilinx" or Nordic?

lib/system/freertos/assert.h
```diff
@@ -0,0 +1,52 @@
...
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Xilinx nor the names of its contributors may be used
```
wjliang: The should the license text be "Xilinx" or Nordic?

lib/system/generic/assert.h
```diff
@@ -0,0 +1,53 @@
...
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Xilinx nor the names of its contributors may be used
```
wjliang: The should the license text be "Xilinx" or Nordic?

lib/system/linux/assert.h
```diff
@@ -0,0 +1,53 @@
...
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Xilinx nor the names of its contributors may be used
```
wjliang: The should the license text be "Xilinx" or Nordic?

lib/system/zephyr/assert.h
```diff
@@ -0,0 +1,53 @@
...
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of Xilinx nor the names of its contributors may be used
```
wjliang: The should the license text be "Xilinx" or Nordic?

wjliang:
there is license text error in the commits, please also fix the license text in lib/system/zephyr/log.h and lib/system/zephyr/log.c

lib/system/zephyr/assert.h
wjliang: I think it better to change, otherwise, "Xilinx" should be in the copyright text.

skordal:
All license issues fixed.

lib/assert.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) 2018, Xilinx Inc. and Contributors. All rights reserved.
```
wjliang: You will need to also add Nordic to the Copyright

wjliang:
commit by split the change to zephyr/log.c|h to another commit

PR #40, opened 2018-04-07
Covert license text to use SPDX tags

wjliang:
Replace the license text with SPDX tags


wjliang:
merged

PR #41, opened 2018-04-07
io: add names property of metal I/O region

wjliang:
Each I/O region can have a name, so that the user can know
what the I/O region is for.

This will line up with carve out memory in the resource table. In the OpenAMP implementation, each remoteproc can have a libmetal device, and each device manage its register I/O regions and shared memory I/O regions.


galak:
I think we should hold on this til, we figure out what the new memory interfaces might look like from the openamp side.

PR #42, opened 2018-04-09
rework irqs_ init to reduce code size

galak:
Rework how we initialize metal_irqs_state _irqs so we can remove metal_irq_init/metal_irq_deinit and reduce code size when the irq register code isn't used.

### DIFF
--- a/lib/system/zephyr/mutex.h
+++ b/lib/system/zephyr/mutex.h
@@ -25,6 +25,15 @@ extern "C" {
 
 typedef struct k_sem metal_mutex_t;
 
+/*
+ * METAL_MUTEX_INIT - used for initializing an mutex elmenet in a static struct
+ * or global
+ */
+#define METAL_MUTEX_INIT(m) _K_SEM_INITIALIZER(m, 1, 1)
+/*
+ * METAL_MUTEX_DEFINE - used for defining and initializing a global or
+ * static singleton mutex
+ */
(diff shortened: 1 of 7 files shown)


lib/system/zephyr/mutex.h
```diff
@@ -49,6 +49,7 @@ extern "C" {
 
 typedef struct k_sem metal_mutex_t;
 
+#define METAL_MUTEX_INIT(m) _K_SEM_INITIALIZER(m, 1, 1)
 #define METAL_MUTEX_DEFINE(m) K_SEM_DEFINE(m, 1, 1)
```
wjliang: Just want to know what's the difference METAL_MUTEX_INIT and METAL_MUTEX_DEFINE?

galak: METAL_MUTEX_INIT is for when we have a struct element we need to initialize.  META_MUTEX_DEFINE is for a singleton global or static.  I've updated the comments and commit messages that will hopefully help.

galak:
Update the commit messages to hopefully clarify difference between INIT/DEFINE & INIT/DECLARE for mutex & lists.  Also, added some comments in the code.

wjliang:
merged with modification to change the mutex define in linux to use mutex init

galak:
> merged with modification to change the mutex define in linux to use mutex init

oops, thanks for fixing that up.

PR #43, opened 2018-04-10
io: use memcpy()/memset() for default block ops

skordal:
Changes the implementations of `metal_io_block_read()`/`_write()`/`_set()` to use `memcpy()` and `memset()`. The original implementation duplicates these functions, and removing them saves 122 bytes when compiling for ARM.


skordal:
I think it's safe to close this now.

PR #45, opened 2018-04-11
zephyr: use sys_cache_flush() for flushing caches

skordal:
Implements the __metal_cache_flush() function as a call to Zephyr's
sys_cache_flush() function.

Signed-off-by: Kristian Klomsten Skordal <kristian.skordal@nordicsemi.no>


wjliang:
merged

PR #46, opened 2018-05-02
[RFC] Events

wjliang:
Introduce events API. User can use the events API to post an event, and register handlers to handle the events.
IRQ and polling can be built on top of this event API.


lib/event.c
```diff
@@ -0,0 +1,92 @@
...
+	event->hd.arg = NULL;
+	metal_spinlock_release(&event->slock);
+}
+
+int metal_event_post(struct metal_event *event)
+{
+	int ret;
+
+	metal_assert(event != NULL);
+	metal_spinlock_acquire(&event->slock);
```
wjliang: will need to disable interrupt before acquire a lock

lib/event.h
```diff
@@ -0,0 +1,214 @@
...
+
+/** \defgroup event Event Handling Interfaces
+ *  @{ */
+
+#include <metal/compiler.h>
+#include <metal/list.h>
+#include <metal/spinlock.h>
+
+/** Event states */
+#define METAL_EVENT_IDLE        0x0UL
```
wjliang: Please explain what each state mean and the transaction between states.

lib/event.c
```diff
@@ -0,0 +1,92 @@
...
+	event->state |= METAL_EVENT_ENABLED;
+	if (event->state & METAL_EVENT_PENDING)
+		event->state &= ~METAL_EVENT_PENDING;
+	metal_spinlock_release(&event->slock);
+	metal_event_post(event);
+}
+
+void metal_event_disable(struct metal_event *event)
+{
```
wjliang: does disable will discard the events? E.g. if there is an interrupt of inter-processor communication raised while the inter-processor communication event is disabled, shall the event marked as PENDING or ignore it?
How about have EVENT_HANDLER_ENABLED state?

wjliang:
Please consider software timer events abstraction in libmetal

lib/event.c
willw-xilinx: Would be useful to be able to discard any pending and queued events.

PR #51, opened 2018-05-22
zephyr: Introduce WITH_ZEPHYR_LIB option

galak:
Introduce an option for when we build libmetal as part of Zephyr itself
instead of as a standalone library.  This is for when libmetal is part
of the Zephyr code base and being built as part of a standard Zephyr
application.  The standalone (WITH_ZEPHYR_LIB=OFF) case we first build
zephyr and use that version to build a libmetal.a and associated
headers.


wjliang:
Does it mean there will be a copy of libmetal side Zephyr? just want to know is there other lib similar to libmetal situation in Zephyr app, do they also get a copy in the Zephyr source code?

galak:
Yes, Zephyr embedded a number of library in the ext/ dir like SoC HALs, crypto libs (mbedtls, tinycrypt), file system libraries, etc:

https://github.com/zephyrproject-rtos/zephyr/tree/master/ext

PR #52, opened 2018-05-22
cmake: Only look for doxygen package if docs enabled

galak:
Add an if (WITH_DOCS) wrapper around find_package(Doxygen).  We only
need Doxygen if we are trying to build the docs.


wjliang:
why don't build the doxygen if doxygen package is found?

PR #53, opened 2018-06-26
fix miscellaneous compilation issue

arnop2:
This patch set is a fix of compilation issue faced using IAR IDE for cortex M.
-  return for void function
- void pointer incrementation
- default implementation of the 64 bit atomic access for 32 bit arm core

### DIFF
--- a/lib/io.c
+++ b/lib/io.c
@@ -33,7 +33,8 @@ void metal_io_init(struct metal_io_region *io, void *virt,
 int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 	       void *restrict dst, int len)
 {
-	void *ptr = metal_io_virt(io, offset);
+	unsigned char *ptr = metal_io_virt(io, offset);
+	unsigned char *dest = dst;
 	int retlen;
 
 	if (offset > io->size)
@@ -47,20 +48,20 @@ int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 	} else {
 		atomic_thread_fence(memory_order_seq_cst);
 		while ( len && (
-			((uintptr_t)dst % sizeof(int)) ||
+			((uintptr_t)dest % sizeof(int)) ||
 			((uintptr_t)ptr % sizeof(int)))) {
-			*(unsigned char *)dst =
+			*(unsigned char *)dest =
 				*(const unsigned char *)ptr;
-			dst++;
+			dest++;
 			ptr++;
 			len--;
 		}
-		for (; len >= (int)sizeof(int); dst += sizeof(int),
+		for (; len >= (int)sizeof(int); dest += sizeof(int),
 					ptr += sizeof(int),
 					len -= sizeof(int))
-			*(unsigned int *)dst = *(const unsigned int *)ptr;
-		for (; len != 0; dst++, ptr++, len--)
-			*(unsigned char *)dst =
+			*(unsigned int *)dest = *(const unsigned int *)ptr;
+		for (; len != 0; dest++, ptr++, len--)
+			*(unsigned char *)dest =
 				*(const unsigned char *)ptr;
 	}
 	return retlen;
@@ -69,7 +70,8 @@ int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 int metal_io_block_write(struct metal_io_region *io, unsigned long offset,
 	       const void *restrict src, int len)
 {
-	void *ptr = metal_io_virt(io, offset);
+	unsigned char *ptr = metal_io_virt(io, offset);
+	const unsigned char *source = src;
 	int retlen;
 
 	if (offset > io->size)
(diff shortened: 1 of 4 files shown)


lib/io.c
```diff
@@ -33,7 +33,8 @@ void metal_io_init(struct metal_io_region *io, void *virt,
 int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 	       void *restrict dst, int len)
 {
-	void *ptr = metal_io_virt(io, offset);
+	char *ptr = metal_io_virt(io, offset);
+        char *dest = dst;
```
wjliang: indentation

lib/io.c
```diff
@@ -33,7 +33,8 @@ void metal_io_init(struct metal_io_region *io, void *virt,
 int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 	       void *restrict dst, int len)
 {
-	void *ptr = metal_io_virt(io, offset);
+	char *ptr = metal_io_virt(io, offset);
```
wjliang: unsigned char is also used in this function, please use char or unsigned char consistently inside the funciton

lib/io.c
```diff
@@ -69,7 +70,8 @@ int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 int metal_io_block_write(struct metal_io_region *io, unsigned long offset,
 	       const void *restrict src, int len)
 {
-	void *ptr = metal_io_virt(io, offset);
+	char *ptr = metal_io_virt(io, offset);
+	const char *source = src;
```
wjliang: please use char or unsigned char consistently in the function

lib/io.c
```diff
@@ -105,7 +107,7 @@ int metal_io_block_write(struct metal_io_region *io, unsigned long offset,
 int metal_io_block_set(struct metal_io_region *io, unsigned long offset,
 	       unsigned char value, int len)
 {
-	void *ptr = metal_io_virt(io, offset);
+	char *ptr = metal_io_virt(io, offset);
```
wjliang: please use char consistently in the function

lib/processor/arm/cpu.h
```diff
@@ -14,4 +14,32 @@
...
+
+  atomic_store_explicit((atomic_ulong *)lptr, *((atomic_ulong *)tmp), order);
+  tmp += sizeof(atomic_ulong);
+  lptr += sizeof(atomic_ulong);
+  atomic_store_explicit((atomic_ulong *)lptr, *((atomic_ulong *)tmp), order);
+}
+
+static inline uint64_t metal_processor_io_read64(atomic_ullong *ptr,
+						 memory_order order)
+{
```
wjliang: The same as the previous function, can we use static lock in this function to make read atomic?

lib/processor/arm/cpu.h
```diff
@@ -14,4 +14,32 @@
 
 #define metal_cpu_yield()
 
+static inline void metal_processor_io_write64(atomic_ullong *ptr,
```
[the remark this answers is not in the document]
arnop2: FYI, this code is a copy past from microblaze processor. 
FYI, this code is a copy past from microblaze processor. 
I had hesitated ,  but i did not find something specify that the access should be atomic. i would let it without protection, as it could be added later if needed. 
Please tell me your preference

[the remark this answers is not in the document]
wjliang: The MicroBlaze code will need be updated. I would prefer atomic as this operation is to write 64bit data, if it is not atomic, you cannot guarantee it writes 64bit successfully.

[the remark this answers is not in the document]
arnop2: Lock protection does not solve the issue. lock just protect from a concurent access in 64 bits, but a 32 bit access can be done in parallel. 
looking deeper in code, an alternative solution would be to not implement the 64 access if NO_ATOMIC_64_SUPPORT is set.  Means in io.h :
-	else if (ptr && sizeof(atomic_ullong) == width)
#ifndef NO_ATOMIC_64_SUPPORT
+	else if (ptr && sizeof(atomic_ullong) == width)
		atomic_store_explicit((atomic_ullong *)ptr, value, order);
- #else
-		metal_processor_io_write64((atomic_ullong *)ptr, value, order);
#endif
	else
		metal_assert (0);
As openamp doesn't perform 64 bit access, this seems reasonnable that application directly handles the protection of a 64 bit access, if needed for platform that doers not support  atomic 64 bit access ( for instance microblaze)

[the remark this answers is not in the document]
wjliang: You are right, even if we have lock inside the write/read64 it will not solve race complete as it is possible that there will be write/read32 or others at the same time, even though this is more like an application issue. for now, let's just assert if for write/read64 if NO 64bit  support.

lib/io.h
```diff
@@ -348,7 +343,7 @@ int metal_io_block_write(struct metal_io_region *io, unsigned long offset,
 int metal_io_block_set(struct metal_io_region *io, unsigned long offset,
 	       unsigned char value, int len);
 
-#include <metal/system/@PROJECT_SYSTEM@/io.h>
+#include <metal/system/generic/io.h>
```
wjliang: It should be "<metal/system/@PROJECT_SYSTEM@/io.h>"

arnop2: oops... apologize for this stupid mistake,

PR #54, opened 2018-06-27
lib: system: freertos: mutex: change to use atomic_int

bentheredonethat:
As freertos does not provide way to statically initialize semaphore,
use atomic int which does provide this functionality.

Signed-off-by: Ben Levinsky <ben.levinsky@xilinx.com>


wjliang:
failed FreeRTOS compilation reported from CI, please include atomic header file

PR #57, opened 2018-07-11
RFC: extra fixes IAR compilation issue 

arnop2:
Hello,

FYI, here is 2 additional fixes for IAR compilation, that i did not proposed but needed for IAR compilation. 
From my POV, they can not be merged, in this version, in libmetal...but any suggestion is welcome to integrate a fix for these both issues.

- "Atomic function treating integers or flags"
=> declare a union for the spinlock and mutex to avoid compilation error due to bad parameter type for
atomic_flag_*  functions.

- "Rename duplicated source file" patch
=>This patch is not exhaustive and just fixes issue for my build environment in IAR . A full patch should rename all the functions and modify associated cmake files...

- "compiler: add IAR compiler file" patch 
=> Clean patch already proposed in pull resuest #56 

Best Regards 
Arnaud


lib/system/generic/condition.c
```diff
@@ -25,7 +25,7 @@ int metal_condition_wait(struct metal_condition *cv,
 	if (!cv || !m || !metal_mutex_is_acquired(m))
 		return -EINVAL;
 
-	if (!atomic_compare_exchange_strong(&cv->m, &tmpm, m)) {
+	if (!atomic_compare_exchange_strong(&cv->m->v, &tmpm->v, m->v)) {
```
wjliang: This was to verify the input mutex, it doesn't looks like it is necessary to use atomic operation for this checking.

lib/spinlock.h
```diff
@@ -22,7 +22,10 @@ extern "C" {
 /** \defgroup spinlock Spinlock Interfaces
  *  @{ */
 struct metal_spinlock {
-	atomic_int v;
+	union{
+		atomic_int v;
+		atomic_flag w;
+	};
```
wjliang: the atomic_flag is only available from c11, how about in the lib/atomic.h, we define atomic_flag if it is not defined.

And use atomic_flag only in the mutex and spinlock, as when checking the mutex and spinlock implementation in generic, it only uses flag.

lib/system/generic/generic_io.c
```diff
@@ -15,7 +15,7 @@ void metal_sys_io_mem_map(struct metal_io_region *io)
 {
 	unsigned long p;
 	size_t psize;
-	void *va;
+	char *va;
```
wjliang: This looks like a separate commit

arnop2: Yes, missing it, i will send it in a separate patch

PR #58, opened 2018-07-19
libmetal: Add API to read device tree property

mubinsyed:
Added new API "metal_linux_get_device_property" to read
device tree property from Linux user space.

### DIFF
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -636,3 +636,32 @@ int metal_generic_dev_sys_open(struct metal_device *dev)
 	return 0;
 }
 
+/**
+ * @brief	Read device tree property from sys filesystem.
+ * @param[in]	device	metal_device of the intended DT node
+ * @param[in]	property_name	name of the property to be read
+ * @param[out]	output	output buffer to store read data
+ * @param[in]	len	number of bytes to be read
+ * @return	0 on success, or -errno on error.
+ */
+
+int metal_linux_get_device_property(struct metal_device *device,
+		const char *property_name, void *output, int len)
+{
+	int fd = 0;
+	int status = 0;
+	const int flags = O_RDONLY;
+	const int mode = S_IRUSR | S_IRGRP | S_IROTH;
+	struct linux_device *ldev = to_linux_device(device);
+	char path[PATH_MAX];
+
+	snprintf(path, sizeof(path), "%s/of_node/%s",
+			 ldev->sdev->path, property_name);
+	fd = open(path, flags, mode);
+	if (fd < 0)
+		return -errno;
+	status = read(fd, output, len);
+
+	return status < 0 ? -errno : 0;
+}
+
--- a/lib/system/linux/sys.h
+++ b/lib/system/linux/sys.h
@@ -46,6 +46,8 @@ extern "C" {
 #define METAL_INVALID_VADDR     NULL
 #define MAX_PAGE_SIZES		32
 
+struct metal_device;
+
 /** Structure of shared page or hugepage sized data. */
 struct metal_page_size {
 	/** Page size. */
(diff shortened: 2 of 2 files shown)


wjliang:
> I dont see how this makes sense for libmetal. This is very specific to linux & dt. Which isn't that portable across other OSes/RTOSes.

We can extend this to metal_get_device_property() in future, but for now, it should not be in the lib/device.h but just put in the lib/system/linux

lib/device.h
```diff
@@ -150,6 +150,7 @@ metal_device_io_region(struct metal_device *device, unsigned index)
 		: NULL);
 }
 
+extern int metal_linux_get_device_property(struct metal_device *device, const char *property_name, char *output, int len);
 /** @} */
```
wjliang: please move this declaration lib/system/linux

wjliang: please change "char *" to "void *" as the output data may not be "char *"

mubinsyed: I couldn't  find device.h counterpart for lib/system/linux,  Can i create it and move the declaration there?

wjliang: as this is for Linux only for now. keep it in linux/device.c only, and in the drivers, uses "extern" declaration as we don't expect to expose this API to users yet.

galak:
I'm still confused why this needs to be an API in libmetal at all?

galak:
> Different devices have its own properties, in Linux, it is device tree, for other OS it can be different. We would like to have an abstraction to get the device properties, so that we can write a driver which can be used across different operating systems.

this API seems way to specific, It seems like either having an API to get what you need (ie register addr, irq #, etc) and than implement that w/a DT call for linux.

wjliang:
> this API seems way to specific, It seems like either having an API to get what you need (ie register addr, irq #, etc) and than implement that w/a DT call for linux.

register address we have already abstracted with the I/O regions, and IRQ is already part of the metal device. This is more for device specific properties such as device's fifo-depth.

galak:
> register address we have already abstracted with the I/O regions, and IRQ is already part of the metal device. This is more for device specific properties such as device's fifo-depth.

something like that seems way too specific to try and abstract in libmetal.

wjliang:
> something like that seems way too specific to try and abstract in libmetal.
It like more like search for the value of the property from the property key. It just looks to me that we can have some abstraction to get the device property across different OSes, even the implementation can be different.

lib/system/linux/device.c
```diff
@@ -636,3 +636,36 @@ int metal_generic_dev_sys_open(struct metal_device *dev)
...
+	const int flags = O_RDONLY;
+	const int mode = S_IRUSR | S_IRGRP | S_IROTH;
+	struct linux_device *ldev = to_linux_device(device);;
+	char path [PATH_MAX];
+
+	snprintf(path, sizeof(path), "%s/of_node/%s",ldev->sdev->path,property_name);
+
+	fd = open(path,flags,mode);
+	if (fd < 0)
+	{
```
wjliang: Please remove the "{" and "}" if there is single statement, we will follow the Linux kernel coding style

lib/system/linux/device.c
```diff
@@ -636,3 +636,36 @@ int metal_generic_dev_sys_open(struct metal_device *dev)
...
+ * @param	device		metal_device of the intended
+ *				device tree node
+ * @param[in]	property_name   name of the property to be read
+ * @param[out]	output          output buffer where read data wil be stored 
+ * @param[in]	len             number of bytes to be read                              
+ * @return	0 on success, or -errno on error.
+ */
+
+int metal_linux_get_device_property(struct metal_device *device, const char *property_name, void *output, int len)
+{
```
wjliang: please try to limit each line to 80 characters.

lib/system/linux/device.c
```diff
@@ -636,3 +636,36 @@ int metal_generic_dev_sys_open(struct metal_device *dev)
 	return 0;
 }
 
+/**
+ * @brief	Read device tree property from sys filesystem.
+ * @param	device		metal_device of the intended
+ *				device tree node
+ * @param[in]	property_name   name of the property to be read
+ * @param[out]	output          output buffer where read data wil be stored
```
wjliang: indentation

mubinsyed: Not sure why it is being shown like this in diff, I see indention is proper, when i view file from github.

wjliang: It is due to the mix usage of space and tabl

wjliang:
Merged with changes for indentation and move comments to header file

Issue #59, opened 2018-07-20
failed on PYNQ


ly0:
Running test-metal-shared and get a failed result

```
root@localhost:~/libmetal/build/test# sudo ./test-metal-shared 
libhugetlbfs: WARNING: Unable to find default kernel huge page size
metal: debug:     Failed fread /dev/urandom
metal: debug:     added page size 4096 @/tmp
metal: debug:     registered platform bus
metal: info:      running [atomic]
metal: info:      result [atomic]............................ pass
metal: info:      running [mutex]
metal: info:      result [mutex]............................. pass
metal: info:      running [shmem]
metal: info:      result [shmem]............................. pass
metal: info:      running [condition]
metal: info:      result [condition]......................... pass
metal: info:      running [spinlock]
metal: info:      result [spinlock].......................... pass
metal: info:      running [alloc]
metal: info:      result [alloc]............................. pass
metal: info:      running [irq]
metal: error:     unregister irq 3 match handler success but fail expected
metal: info:      result [irq]............................... fail - error: Invalid argument
metal: info:      running [version]
metal: debug:     metal_linux_irq_handling: poll unexpected. fd 8: 32
metal: info:      result [version]........................... pass
metal: debug:     metal_linux_irq_shutdown
metal: debug:     metal_linux_irq_handling: poll unexpected. fd 8: 32
metal: debug:     unregistered platform bus
```

my dts:

```dts
/ {
    compatible = "xlnx,zynq-7000";

    reserved-memory {
        #address-cells = <1>;
        #size-cells = <1>;
        /* Reserved memory for both firmware and shared memory */
        rproc_0_reserved: rproc@3ed000000 {
            no-map;
            reg = <0x18000000 0x8000000>;
        };
    };

 
    amba {
	 remoteproc0: remoteproc@0 {
	 	compatible = "xlnx,zynq_remoteproc";
	 	reg = <0x00000000 0x10000000>;
	 	firmware = "firmware";
	 	vring0 = <15>;
	 	vring1 = <14>;
	 };
        shm0: shm@0 {
            compatible = "shm_uio";
            reg = <0x0 0x1ed80000 0x0 0x80000>;
        };
 
        ipi0: ipi@0 {
            compatible = "ipi_uio";
            reg = <0x0 0xff340000 0x0 0x1000>;
            interrupt-parent = <&intc>;
            interrupts = <0 29 4>;
        };
 
 
    };
};

```

arnopo:
[...]

PR #64, opened 2018-09-20
fix urandom read seed

tyokota-eforce:
original code always fails size checking from reading /dev/urandom in my enviroment.

>metal: debug: Failed fread /dev/urandom
>metal: debug: added page size 4096 @/tmp
>metal: debug: registered platform bus
>metal: debug: opened sysfs device platform:1ff00000.vring

### DIFF
--- a/lib/system/linux/init.c
+++ b/lib/system/linux/init.c
@@ -135,7 +135,7 @@ int metal_sys_init(const struct metal_init_params *params)
 			  strerror(errno));
 		return -errno;
 	}
-	if (sizeof(int) != fread(&seed, sizeof(int), 1, urandom)) {
+	if (sizeof(unsigned int) != fread(&seed, 1, sizeof(unsigned int), urandom)) {
 		metal_log(METAL_LOG_DEBUG, "Failed fread /dev/urandom\n");
 	}
 	fclose(urandom);


wjliang:
Could you give more details on how your fix will solve the issue on your system?
Is sizeof(unsigned int) different to sizeof(int) in your system? Just it doesn't look like the change will change the number of bytes read to me. "fread(&seed, sizeof(int), 1, urandom)" looks like it  will return the same as "fread(&seed, 1, sizeof(unsigned int), urandom)" to me.

tyokota-eforce:
>Is sizeof(unsigned int) different to sizeof(int) in your system? 

no.it's same size.
main purpose of this code is changing the order of parameter.


>"fread(&seed, sizeof(int), 1, urandom)" looks like it will return the same as "fread(&seed, 1, sizeof(unsigned int), urandom)"

it's different.
first code returns 1. second one returns 4.

wjliang:
> > Is sizeof(unsigned int) different to sizeof(int) in your system?
> 
> no.it's same size.
> main purpose of this code is changing the order of parameter.
> 
> > "fread(&seed, sizeof(int), 1, urandom)" looks like it will return the same as "fread(&seed, 1, sizeof(unsigned int), urandom)"
> 
> it's different.
> first code returns 1. second one returns 4.
OK, I see fread returns number of items it  reads. how about "if (fready(&seed, sizeof(seed), 1, urandom) <= 0)" ?

tyokota-eforce:
I think "if (fready(&seed, 1, sizeof(seed), urandom) <= 0)" is more better.

https://github.com/OpenAMP/libmetal/issues/59 log seems current urandom read, "4 byte size  read", failed on petalinux platform or arm linux.
my enviroment is angstrom linux on intel socfpga.

wjliang:
> I think "if (fread(&seed, 1, sizeof(seed), urandom) <= 0)" is better.

fread() may returns sth like >0 && <sizeof(seed), but /dev/urandom should not have this issue, if it fails, it is likely that the file doesn't exist or permissions. Please go ahead with "if (fread(&seed, 1, sizeof(seed), urandom) <= 0)" as you posted

wjliang:
merged

PR #68, opened 2018-11-06
Hyunk/driver probe

starhwk:
This fixes the issue where there are multiple drivers under same bus.

### DIFF
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -156,6 +156,7 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
 
 
 	ldev->fd = -1;
+	ldev->device.irq_info = (void *)-1;
 
 	ldev->sdev = sysfs_open_device(lbus->bus_name, ldev->dev_name);
 	if (!ldev->sdev) {
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -586,34 +586,32 @@ static int metal_linux_probe_driver(struct linux_bus *lbus,
 	return ldrv->sdrv ? 0 : -ENODEV;
 }
 
+static void metal_linux_bus_close(struct metal_bus *bus);
+
 static int metal_linux_probe_bus(struct linux_bus *lbus)
 {
 	struct linux_driver *ldrv;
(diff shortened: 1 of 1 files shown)


lib/system/linux/device.c
```diff
@@ -447,10 +447,8 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 
 		/* Try and open the device. */
 		error = ldrv->dev_open(lbus, ldev);
-		if (error) {
-			ldrv->dev_close(lbus, ldev);
+		if (error)
 			continue;
```
wjliang: Better to leave it, as it forces the driver dependent device open to clean up properly when it has issues during opening.
E.g. the uio_dev_open() didn't do that.
Or does it introduce any issues?

lib/system/linux/device.c
```diff
@@ -586,34 +586,35 @@ static int metal_linux_probe_driver(struct linux_bus *lbus,
...
 
 	lbus->sbus = sysfs_open_bus(lbus->bus_name);
 	if (!lbus->sbus)
 		return -ENODEV;
 
 	for_each_linux_driver(lbus, ldrv) {
-		error = metal_linux_probe_driver(lbus, ldrv);
-		if (!error)
+		if (!ldrv->drv_name)
 			break;
```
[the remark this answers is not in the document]
starhwk: ah, for_each_linux_driver() already has this check. I didn't see. Let me remove this check.

lib/system/linux/device.c
starhwk: If open fails, that means the device isn't opened. Closing such device doesn't seem straightforward. As an example, posix open/close can be looked as analogy. If any cleanup is needed, better to be done in the error handling in the open call.
In my case, the close is calling metal_irq_unregister() with invalid fd, and it was printing error, even when everything is still operational. So I removed this to avoid misleading.

wjliang: ok, I see, better to have another patch to the existing driver dev_open to make sure it clean up when it fails in the dev_open(). But this can be another pull request.
Please go ahead with your change for this pull request. We can worry about to cleanup properly in the driver's dev_open() later

PR #71, opened 2018-12-24
Upstream from Pinecone

xiaoxiang781216:
@wjliang, these series patch include:
1.the minor bugfix
2.new cpu support
  a.ceva(https://www.ceva-dsp.com/)
  b.csky(http://en.c-sky.com/)
  c.riscv(https://riscv.org/)
3.nuttx system support(http://www.nuttx.org/)
Please review them, thanks.

### DIFF
--- a/lib/system/nuttx/io.h
+++ b/lib/system/nuttx/io.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2018, Pinecone Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	nuttx/io.h
+ * @brief	NuttX specific io definitions.
+ */
+
+#ifndef __METAL_IO__H__
+#error "Include metal/io.h instead of metal/nuttx/io.h"
+#endif
+
+#ifndef __METAL_NUTTX_IO__H__
+#define __METAL_NUTTX_IO__H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief  Get the default global io ops.
+ * @return  an io ops.
+ */
+struct metal_io_ops *metal_io_get_ops(void);
+
+/**
+ * @brief  Get the default global io region.
+ * @return  an io region.
+ */
+struct metal_io_region *metal_io_get_region(void);
+
+#ifdef METAL_INTERNAL
+
+/**
+ * @brief memory mapping for an I/O region
+ */
+static inline void metal_sys_io_mem_map(struct metal_io_region *io)
+{
+}
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __METAL_NUTTX_IO__H__ */
--- a/lib/system/nuttx/irq.c
+++ b/lib/system/nuttx/irq.c
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2018, Pinecone Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	nuttx/irq.c
+ * @brief	NuttX libmetal irq definitions.
+ */
+
+#include <errno.h>
+#include <metal/irq.h>
+#include <metal/alloc.h>
+#include <nuttx/irq.h>
+
+/** IRQ handler descriptor structure */
+struct metal_irq_hddesc {
+	metal_irq_handler hd;     /**< irq handler */
+	void *drv_id;             /**< id to identify the driver
+	                               of the irq handler */
+};
+
+static int metal_irq_isr(int irq, void *context, void *arg);
+
+int metal_irq_register(int irq,
+		       metal_irq_handler hd,
+		       struct metal_device *dev,
+		       void *drv_id)
+{
+	struct metal_irq_hddesc *desc;
+	int ret;
+
+	if ((drv_id == NULL) || (hd == NULL))
+		return -EINVAL;
+
+	desc = metal_allocate_memory(sizeof(struct metal_irq_hddesc));
+	if (desc == NULL)
+		return -ENOMEM;
+
+	desc->hd = hd;
+	desc->drv_id = drv_id;
+
+	ret = irq_attach(irq, metal_irq_isr, desc);
+	if (ret < 0)
+		metal_free_memory(desc);
+
+	return ret;
+}
+
+int metal_irq_unregister(int irq,
+			 metal_irq_handler hd,
+			 struct metal_device *dev,
+			 void *drv_id)
+{
+	unsigned int irq_flags_save;
+
+	if (irq < 0)
+		return -EINVAL;
+
+	irq_flags_save = metal_irq_save_disable();
+	/* context == drv_id mean unregister */
+	irq_dispatch(irq, drv_id); /* fake a irq request */
+	metal_irq_restore_enable(irq_flags_save);
+
+	return 0;
+}
+
+unsigned int metal_irq_save_disable(void)
+{
+	return enter_critical_section();
+}
+
+void metal_irq_restore_enable(unsigned int flags)
+{
+	leave_critical_section(flags);
+}
+
+void metal_irq_enable(unsigned int vector)
+{
+	up_enable_irq(vector);
+}
+
+void metal_irq_disable(unsigned int vector)
+{
+	up_disable_irq(vector);
+}
+
+/**
+  * @brief       IRQ handler
+ */
+static int metal_irq_isr(int irq, void *context, void *arg)
+{
+	struct metal_irq_hddesc *desc = arg;
+
+	if (context != desc->drv_id)
+		return desc->hd(irq, desc->drv_id);
+
+	/* context == drv_id mean unregister */
+	irqchain_detach(irq, metal_irq_isr, arg);
+	sched_kfree(arg);
+	return 0;
+}
--- a/lib/system/nuttx/CMakeLists.txt
+++ b/lib/system/nuttx/CMakeLists.txt
@@ -1,0 +1,20 @@
+collect (PROJECT_LIB_HEADERS alloc.h)
+collect (PROJECT_LIB_HEADERS assert.h)
+collect (PROJECT_LIB_HEADERS cache.h)
+collect (PROJECT_LIB_HEADERS condition.h)
+collect (PROJECT_LIB_HEADERS io.h)
+collect (PROJECT_LIB_HEADERS irq.h)
+collect (PROJECT_LIB_HEADERS log.h)
+collect (PROJECT_LIB_HEADERS mutex.h)
+collect (PROJECT_LIB_HEADERS sleep.h)
+collect (PROJECT_LIB_HEADERS sys.h)
+
+collect (PROJECT_LIB_SOURCES condition.c)
+collect (PROJECT_LIB_SOURCES device.c)
+collect (PROJECT_LIB_SOURCES init.c)
+collect (PROJECT_LIB_SOURCES io.c)
+collect (PROJECT_LIB_SOURCES irq.c)
+collect (PROJECT_LIB_SOURCES shmem.c)
+collect (PROJECT_LIB_SOURCES time.c)
+
+# vim: expandtab:ts=2:sw=2:smartindent
--- a/lib/system/nuttx/alloc.h
+++ b/lib/system/nuttx/alloc.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2018, Pinecone Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	nuttx/alloc.h
+ * @brief	nuttx libmetal memory allocattion definitions.
+ */
+
+#ifndef __METAL_ALLOC__H__
+#error "Include metal/alloc.h instead of metal/nuttx/alloc.h"
+#endif
+
+#ifndef __METAL_NUTTX_ALLOC__H__
+#define __METAL_NUTTX_ALLOC__H__
+
+#include <nuttx/kmalloc.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline void *metal_allocate_memory(unsigned int size)
+{
+	return kmm_malloc(size);
+}
+
+static inline void metal_free_memory(void *ptr)
+{
+	kmm_free(ptr);
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __METAL_NUTTX_ALLOC__H__ */
--- a/lib/system/nuttx/assert.h
+++ b/lib/system/nuttx/assert.h
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2018, Pinecone Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	assert.h
+ * @brief	NuttX assertion support.
+ */
+
+#ifndef __METAL_ASSERT__H__
+#error "Include metal/assert.h instead of metal/nuttx/assert.h"
+#endif
+
+#ifndef __METAL_NUTTX_ASSERT__H__
+#define __METAL_NUTTX_ASSERT__H__
+
+#include <assert.h>
+
+/**
+ * @brief Assertion macro for NuttX-based applications.
+ * @param cond Condition to evaluate.
+ */
+#define metal_sys_assert(cond) DEBUGASSERT(cond)
+
+#endif /* __METAL_NUTTX_ASSERT__H__ */
--- a/lib/system/nuttx/cache.h
+++ b/lib/system/nuttx/cache.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2018, Pinecone Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	nuttx/cache.h
+ * @brief	NuttX cache operation primitives for libmetal.
+ */
+
+#ifndef __METAL_CACHE__H__
+#error "Include metal/cache.h instead of metal/nuttx/cache.h"
+#endif
+
+#ifndef __METAL_NUTTX_CACHE__H__
+#define __METAL_NUTTX_CACHE__H__
+
+#include <nuttx/arch.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline void __metal_cache_flush(void *addr, unsigned int len)
+{
+	up_clean_dcache((uintptr_t)addr, (uintptr_t)addr + len);
+}
+
+static inline void __metal_cache_invalidate(void *addr, unsigned int len)
+{
+	up_invalidate_dcache((uintptr_t)addr, (uintptr_t)addr + len);
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* __METAL_NUTTX_CACHE__H__ */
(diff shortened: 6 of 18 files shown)


lib/io.h
```diff
@@ -62,6 +62,15 @@ struct metal_io_ops {
...
 	void		(*close)(struct metal_io_region *io);
+	void*		(*offset_to_virt)(struct metal_io_region *io,
+					  unsigned long offset);
+	unsigned long	(*virt_to_offset)(struct metal_io_region *io,
+					  void* virt);
+	metal_phys_addr_t
+			(*offset_to_phys)(struct metal_io_region *io,
+					  unsigned long offset);
+	unsigned long	(*phys_to_offset)(struct metal_io_region *io,
+					  metal_phys_addr_t phys);
```
wjliang: Could you share the use case on why it requires customised offset to virt/phys conversion?

lib/processor/ceva/cpu.h
```diff
@@ -0,0 +1,21 @@
...
+ * @brief	CPU specific primatives
+ */
+
+#ifndef __METAL_CEVA_CPU__H__
+#define __METAL_CEVA_CPU__H__
+
+#define metal_cpu_yield()
+
+#ifdef TEAKLITE4
+#define __sync_synchronize()
```
wjliang: Is this function used to flush cache? could you explain what is this function is used for?

xiaoxiang781216: Since the compiler for CEVA' TEAKLITE4 series DSP doesn't implement __sync_synchronize builtin function, I have to define a dummy here to avoid the linker error.

lib/io.h
xiaoxiang781216: Some complex RTOS(e.g. NuttX) has the dedicated subsystem to manage the memory, which:
1.Encapsulate the hardware MMU difference and
2.Expose a set of API to manage the virtual and physical address mapping
So when OpenAMP is ported to this type of RTOS, it's better to reuse the address conversion API provided by RTOS directly, instead duplicating the translation table again.
The callback is a simple way to achieve this flexibility and keep the backward compatibility at the same time.

lib/processor/ceva/cpu.h
wjliang: just want to know if the compiler doesn't implement this function, is the application trying to use it?
If yes, what is this function do? Usually who define this function?

lib/io.h
wjliang: Here is what's not clear to me: the io region assumes that the  virtual is contiguous but the physical is not, the offset is to the start of the virtual address. In this case, the physmap is an array, you can have multiple phys regions for one IO region. Is there any reason you cannot use physmap array to solve your issue?

Thanks,
Wendy

lib/processor/ceva/cpu.h
xiaoxiang781216: Yes, function in lib/compiler/gcc/atomic.h(e.g. atomic_store, atomic_load, atomic_thread_fence...) call __sync_synchronize.
Normally, atomic function provide by toolchain vendor and put into libgcc.a.
But, since CEVA TL42x DSP doesn't support the out of order execution and the mulitcore architecture, __sync_synchronize doesn't have any meaning on TL42x and CEVA forget to provide a dummy implementation here to improve the compatibility.

lib/io.h
xiaoxiang781216: Yes, physmap array could solve this problem theoretically, but this require:
1.Expose OS internal structure which isn't good from encapsulation, and OS desginer will challenge me why not use the OS API to do the address conversion directly.
2. Some DSP need the very specific VA to PA mapping which can't represent by physmap array. For example, TL42x(many audio DSP) use word(2 byte) as the base unit, DSP address 0x1000 really point to 0x2000 byte, here we have to multiple 2 for va->pa convesion. With callback function, we can avoid this ugly code in the common code. 
3.Finally, the new callback functions follow other callback design(write/read/block_write/block_read/block_set), I think this is a natural extension.

wjliang: > Yes, physmap array could solve this problem theoretically, but this require:
> 1.Expose OS internal structure which isn't good from encapsulation, and OS desginer will challenge me why not use the OS API to do the address conversion directly.

Agreed
> 2. Some DSP need the very specific VA to PA mapping which can't represent by physmap array. For example, TL42x(many audio DSP) use word(2 byte) as the base unit, DSP address 0x1000 really point to 0x2000 byte, here we have to multiple 2 for va->pa convesion. With callback function, we can avoid this ugly code in the common code.

Try to understand more on this point. your address is 2byptes.
when you get the physical address out of his (DSP address 0x1000 and TL42x 0x2000) region.
do you use this physical address locally or you use it remotely? that is do you need to pass this address to the remote?

We also have remote address which is different to local bus address cases.
Is this one causing the issue?

Is your OpenAMP application runs on TL42x core? if yes, what's the physmap set in your application, dsp (0x1000) or 0x2000?

do you use both "0x1000" and "0x2000" in the same application runs on TL42x core?
Have you also consider two io regions in this case?

> 3.Finally, the new callback functions follow other callback design(write/read/block_write/block_read/block_set), I think this is a natural extension.

lib/processor/ceva/cpu.h
wjliang: As your DSP doesn't support out of order execution and multicore, looks like it is atomic still work with this dummy implementation. Looks fine for me, just better to have some comments about this definition.

lib/io.h
xiaoxiang781216: > > Yes, physmap array could solve this problem theoretically, but this require:
> > 1.Expose OS internal structure which isn't good from encapsulation, and OS desginer will challenge me why not use the OS API to do the address conversion directly.
> 
> Agreed
> 
> > 1. Some DSP need the very specific VA to PA mapping which can't represent by physmap array. For example, TL42x(many audio DSP) use word(2 byte) as the base unit, DSP address 0x1000 really point to 0x2000 byte, here we have to multiple 2 for va->pa convesion. With callback function, we can avoid this ugly code in the common code.
> 
> Try to understand more on this point. your address is 2byptes.
> when you get the physical address out of this (DSP address 0x1000 / 0x2000) region.
> do you use this physical address locally or you use it remotely? that is do you need to pass this address to the remote?
> 
TL42x use 0x1000(VA) locally, but it has to convert to 0x2000(PA) before pass to other side.

> We also have remote address which is different to local bus address cases.
> Is this one causing the issue?
This is another issue we face could resolve by physmap,  but TL42x can't use physmap since the mapping isn't linear. Actually, we extend the virtual address concept:
1.Not only include the address which go through MMU
2.but also include the address go through some bus mapping
Both address need go through io_region va<->pa conversion(either physmap or callback).

> In this case, we handle this mapping out of io_region, we manage da_to_pa/pa_to_da within remoteproc in OpenAMP. If this is the case, we had some discussion before on where to put this mapping, in the end, we put this mapping within remoteproc
> 
As I understand, we need three type of address here: VA, PA and DA:
The first pair VA/PA both come from the same CPU, use in the normal IPC comunication, the conversion is provided by io_region.
The second pair PA/DA is different since DA come from another CPU, but PA come from local, this pair is only used during the elf loading and is provided through remoteproc_mem.
I think both mapping VA<->PA and PA<->DA are needed.

> Is your OpenAMP application runs on TL42x core? if yes, what's the physmap set in your application, dsp (0x1000) or 0x2000?
> 
Yes, OpenAMP run on TL42x core. Conceptly, I set VA to 0x1000 and PA to 0x2000, but I really utilize callback here do the mapping since the table can't express the double operation.

> do you use both "0x1000" and "0x2000" in the same application runs on TL42x core?
> 
No, the most code on TL42x use 0x1000, 0x2000 is only used to set the vring buffer address in the share memory.

> > 3.Finally, the new callback functions follow other callback design(write/read/block_write/block_read/block_set), I think this is a natural extension.

lib/processor/ceva/cpu.h
xiaoxiang781216: Ok, I will add some comment here.

arnop2:
The need seems reasonable to me. My only concern here is that this patch increases the libmetal lib footprint that is already high for platforms that have memory size constraint and no memory translation need. 
But i don't have alternative to propose today for this ...
And this is another subject already discussed with wjliang and galak. The libmetal would need a global restructuring to have a better optimization depending on platform and OS.

lib/io.h
wjliang: > linear. Actually, we extend the virtual address concept:
> 1.Not only include the address which go through MMU
> 2.but also include the address go through some bus mapping
> Both address need go through io_region va<->pa conversion(either physmap or callback).

You have virtual to offset and offset to virtual besides physical address. That is for you io_region, even the logical address va is not linear. Could you clarify that? Do you also need offset <-> va mapping?

xiaoxiang781216: No, we don't really need offset<->va mapping. Actually, in NuttX porting:
https://github.com/OpenAMP/libmetal/blob/fd213750017660437f440dad3ecb13eded4e3f19/lib/system/nuttx/io.c
you can see the offset<->va is the identity mapping.
this issue is related to the current io_region desgin:
1.va<->pa mapping need convert to offset first
2.read/write/set use offset notation, not virtual address
I have to follow it.
Actually, I think the offset in io_region is redundancy, if we remove offset from io_region and:
1.convert pa->va and va->pa directly
2.read/wreite/set use va(void *) as input
This desgin is much simple than before.

xiaoxiang781216:
@arnop2 if the footprint is a concern, we can:
1.change metal_io_region ops in metal_io_region to const metal_io_region *ops
2.Move the default table base implementation to callback style, the related variable also could remove from metal_io_region
3.Add a new identity mapping callback implementation
So the OS/platform could select the best implementation, or implement the own translation if the table mapping and identity mapping isn't suitable.

lib/io.h
wjliang: > you can see the offset<->va is the identity mapping.
> this issue is related to the current io_region desgin:
> 1.va<->pa mapping need convert to offset first
But shouldn't the offset is the offset  to the base of the va address, should it be just linear?

You just need to do offset <->pa?
> 2.read/write/set use offset notation, not virtual address
I think the offset is the offset to the base of va, and thus, the offset <-> va should not need any changes.

> I have to follow it.
> Actually, I think the offset in io_region is redundancy, if we remove offset from io_region and:
> 1.convert pa->va and va->pa directly
> 2.read/wreite/set use va(void *) as input
> This desgin is much simple than before.

wjliang:
> 1.change metal_io_region ops in metal_io_region to const metal_io_region *ops
> 2.Move the default table base implementation to callback style, the related variable also could remove from metal_io_region
> 3.Add a new identity mapping callback implementation
> So the OS/platform could select the best implementation, or implement the own translation if the table mapping and identity mapping isn't suitable.

I not clear on your bullet 3, but io.h implementation can the default implementation and allow different OS to provide their own implementation.
A simple way is we can have macro to define if the OS has its own implementation, if no, compile with the default implementation, other wise, compile with the OS implementation

xiaoxiang781216:
> > 1.change metal_io_region ops in metal_io_region to const metal_io_region *ops
> > 2.Move the default table base implementation to callback style, the related variable also could remove from metal_io_region
> > 3.Add a new identity mapping callback implementation
> > So the OS/platform could select the best implementation, or implement the own translation if the table mapping and identity mapping isn't suitable.
> 
> I not clear on your bullet 3, but io.h implementation can the default implementation and allow different OS to provide their own implementation.
> A simple way is we can have macro to define if the OS has its own implementation, if no, compile with the default implementation, other wise, compile with the OS implementation
Yes, the current implementation in io.h could be default. But, even OS/platform don't like the default implemenation and replace with the new callback, this filed in io_region still exist and consume memory.
Since this thread talk about the footprint, let's calculate the size reduction:
1.The current io_region is 48 bytes
2.With my change extend to 64 bytes
3.Change ops to pointer reduce to 28 bytes
  Yes, after the chagne, we need define ops(40 bytes) additionally. But one ops instance is suitable for one OS/platform normally, so multiple io_region could share one ops.
4.Split the table implementation, io_region just become to 4 bytes, now we have:
  1.io_region just contain one field ops pointer
  2.io_table_region contain 28 bytes, one ops 40 bytes
  3.io_identity_region contain 4 bytes, one ops 40 bytes
The current desgin even the platform just need identity mapping, each io_region still consume 28 bytes even after apply the optimization I just mention(or 64 bytes without any optimization).
The new desgin each io_identity_region just 4 bytes.

lib/system/nuttx/io.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+struct metal_io_region *metal_io_get_region(void);
+
+/**
+ * @brief  Convert a physical address to virtual address.
+ * @param[in]  phys  Physical address within segment.
+ * @return  NULL if out of range, or corresponding virtual address.
+ */
+void *metal_io_patova(metal_phys_addr_t phys);
+
```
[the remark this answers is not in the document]
xiaoxiang781216: Ok, I will remove this

lib/io.h
xiaoxiang781216: You just need to do offset <->pa?
No, the really need is va <->pa. As I emphasize before, the current io_region desgin require:
1.va need first convert to offset
2.offset convert to pa(or call read/write/set)
This is the io_region desgin issue, before io_region remove offset notation from all API/callback, the callback for offset<->va and offset<->pa is required.
Once offset notation is removed, we just need add va<->pa callback.

xiaoxiang781216: One more comment:
OS(include NuttX) normally provide va<->pa conversion API directly, not va<->offset and offset<->pa conversion.
To work with io_region offset notation, I have to:
1.va<->offset use the identity mapping
2.offset<->pa call OS va<->pa API

wjliang:
> I not clear on your bullet 3, but io.h implementation can the default implementation and allow different OS to provide their own implementation.

That is to figure out a way to have a default common implementation, and allow OS/platform to have their own implementation if the common one doesn't fit. but this will not be enough if one application have different types of io regions. And I think this can be done in another pull request. Please go ahead to update the pull request just removing the offset<->va

xiaoxiang781216:
> > I not clear on your bullet 3, but io.h implementation can the default implementation and allow different OS to provide their own implementation.
> 
> That is to figure out a way to have a default common implementation, and allow OS/platform to have their own implementation if the common one doesn't fit. but this will not be enough if one application have different types of io regions.

If OS/platform need multiple implementation, it could call e.g.:
1.metal_io_table_init for table mapping
2.metal_io_identity_init for identity mapping
3.metal_io_xxx_init for xxx mapping
So the solution is enough.

>And I think this can be done in another pull request.

Yes, I agree, if this optimation is really needed, it should go to another patch.

> Please go ahead to update the pull request just removing the offset<->va

But, before we remove the offset concept from metal_io_region and convert va<->pa directly, the offset<->va callback make the implemantation of metal_io_virt/metal_io_virt_to_offset more easier.
How we implement these two API without offset<->va callback naturally? So I think we can keep the current desgin until the offset concetp get removed from metal_io_region.

lib/io.h
wjliang: libmetal has metal_io_virt_to_phys() API which sounds like can keep the common implementation but allow OS specific implementation if required.

Just for your io_region, as offset is the offset to va, and thus, i think in your offset <->phys conversion, you can do NuttX_va_to_pa(io->va + offset), and offset = NuttX_pa_to_va(pa) - va.
Will it work?

wjliang:
> But, before we remove the offset concept from metal_io_region and convert va<->pa directly, the offset<->va callback make the implemantation of metal_io_virt/metal_io_virt_to_offset more easier.
> How we implement these two API without offset<->va callback naturally? So I think we can keep the current desgin until the offset concetp get removed from metal_io_region.

Please see this reply:
`
Just for your io_region, as offset is the offset to va, and thus, i think in your offset <->phys conversion, you can do NuttX_va_to_pa(io->va + offset), and offset = NuttX_pa_to_va(pa) - va.
Will it work?
`

lib/io.h
xiaoxiang781216: > libmetal has metal_io_virt_to_phys() API which sounds like can keep the common implementation but allow OS specific implementation if required.

Yes, but it is concrete implementation now, I can't easliy replace with my implemenation.
Unless it change from:
return metal_io_phys(io, metal_io_virt_to_offset(io, virt));
to:
return (*io->ops.virt_to_phys)(io, virt);

> 
> Just for your io_region, as offset is the offset to va, and thus, i think in your offset <->phys conversion, you can do NuttX_va_to_pa(io->va + offset), and offset = NuttX_pa_to_va(pa) - va.
> Will it work?

I can't understand very well the question.
In my case, offset isn't the offset to va, offset actually equal va. So:
1.For va->offset, I just return va as offset
2.For offset->pa, I just call NuttX_va_to_pa(offset)
3.For pa->offset, I just call NuttX_pa_to_va(pa)
4.For offset->va, I just reutrn offset as va

After some thought, I can initialize io->virt=0, io->size=0xffffffff to make metal_io_virt/metal_io_virt_to_offset become the identity mapping, and then remove offset_to_virt/virt_to_offset callback.
Is it your want?

xiaoxiang781216:
@wjliang the new patch remove offset<->va mapping, and achieve identity mapping by io->virt =0 and io->size = 0xffffffff

lib/system/nuttx/irq.c
```diff
@@ -0,0 +1,103 @@
...
+	void *drv_id;             /**< id to identify the driver
+	                               of the irq handler */
+};
+
+static int metal_irq_isr(int irq, void *context, void *arg);
+
+int metal_irq_register(int irq,
+		       metal_irq_handler hd,
+		       struct metal_device *dev,
+		       void *drv_id)
```
wjliang: could you take a look at this pull request: https://github.com/OpenAMP/libmetal/pull/72?
It will introduce a change in the IRQ implementation, which will introduce interrupt controller.
Here is the branch for the change: https://github.com/wjliang/libmetal/tree/master-soft-irq
Please check if it will have a big impact in your implementation.

Thanks,
Wendy

xiaoxiang781216: Ok, I will study the patch in the weekend and summary the impact.

lib/system/nuttx/irq.c
```diff
@@ -0,0 +1,103 @@
...
+		return -EINVAL;
+
+	desc = metal_allocate_memory(sizeof(struct metal_irq_hddesc));
+	if (desc == NULL)
+		return -ENOMEM;
+
+	desc->hd = hd;
+	desc->drv_id = drv_id;
+
+	ret = irq_attach(irq, metal_irq_isr, desc);
```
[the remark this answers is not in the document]
xiaoxiang781216: Here is the free process:
1.metal_irq_unregister trigger a fake interrupt by irq_dispatch
2.metal_irq_isr found that it's a fake interrupt and call irqchain_detach and sched_kfree
It's a little bit hack, since NuttX:
1.Already support the irq chain concept, I want to utilize it to save the coding
2.Three argument(irq, callback and arg) for irqchain_detach need exactly match irq_attach, so irq subsystem could just remove the right one from the list.
3.Since in metal_irq_register, I dynamiclly allocate metal_irq_hddesc, the normally approach could save it in a list and then we can find it in metal_irq_unregister.
4.But this approach make the implementation complicated and waste memory, so I choice a more clever(or hack) approach:
1.Trigger a fake interrupt by software
2.Irq subsystem find the right metal_irq_hddesc
3.Pass the descriptor to metal_irq_isr
4.metal_irq_isr call irqchain_detach with all right information
5.Free the descriptor by sched_kfree

lib/system/nuttx/irq.c
xiaoxiang781216: It's simple to port to the new desgin, but we need improve the desgin to avoid the handler array duplication if RTOS already have the similiar irq dispatch table.

lib/system/nuttx/irq.c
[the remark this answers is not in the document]
xiaoxiang781216: Sure, the irq dispatcher don't check the enable bit when dispatch the incoming interrupt(either real or fake).

wjliang:
this patch breaks the existing Xilinx Zynq UltraScale+ MPSoC demo. still trying to see why

PR #72, opened 2019-01-03
Updated libmetal IRQ impelmentation

wjliang:
Update libmetal IRQ implementation.

The libmetal IRQ abstraction is updated to allow user to implement their own interrupt controller with IRQ abstraction. The driver/application IRQ registration from standalone or userspace can be independent to IRQ controller implementation.

v3:
 * move irq controller related definitions to irq_controller.h
 * allows irq controller to have their own irq registration callback.

v2:
 * Remove pending states from irq.h
 * move soft irq functions to softirq.[ch]
 * remove locking in the registration related functions assuming registration will not be called by multiple threads.

### DIFF
--- a/lib/system/nuttx/init.c
+++ b/lib/system/nuttx/init.c
@@ -10,13 +10,17 @@
  */
 
 #include <metal/device.h>
+#include <metal/irq.h>
 #include <metal/sys.h>
 
 struct metal_state _metal;
 
 int metal_sys_init(const struct metal_init_params *params)
 {
-	return metal_bus_register(&metal_generic_bus);
+	int ret = metal_cntr_irq_init();
+	if (ret >= 0)
+		ret = metal_bus_register(&metal_generic_bus);
+	return ret;
 }
 
 void metal_sys_finish(void)
--- a/lib/system/nuttx/irq.c
+++ b/lib/system/nuttx/irq.c
@@ -10,94 +10,78 @@
  */
 
 #include <errno.h>
-#include <metal/irq.h>
+#include <metal/irq_controller.h>
 #include <metal/alloc.h>
 #include <nuttx/irq.h>
 
-/** IRQ handler descriptor structure */
-struct metal_irq_hddesc {
-	metal_irq_handler hd;     /**< irq handler */
-	void *drv_id;             /**< id to identify the driver
-	                               of the irq handler */
-};
-
-static int metal_irq_isr(int irq, void *context, void *arg);
-
-int metal_irq_register(int irq,
-		       metal_irq_handler hd,
-		       struct metal_device *dev,
-		       void *drv_id)
-{
-	struct metal_irq_hddesc *desc;
-	int ret;
-
-	if ((drv_id == NULL) || (hd == NULL))
-		return -EINVAL;
-
-	desc = metal_allocate_memory(sizeof(struct metal_irq_hddesc));
-	if (desc == NULL)
-		return -ENOMEM;
-
-	desc->hd = hd;
-	desc->drv_id = drv_id;
-
-	ret = irq_attach(irq, metal_irq_isr, desc);
-	if (ret < 0)
-		metal_free_memory(desc);
-
-	return ret;
-}
-
-int metal_irq_unregister(int irq,
-			 metal_irq_handler hd,
-			 struct metal_device *dev,
-			 void *drv_id)
-{
-	unsigned int irq_flags_save;
-
-	if (irq < 0)
-		return -EINVAL;
-
-	irq_flags_save = metal_irq_save_disable();
-	/* context == drv_id mean unregister */
-	irq_dispatch(irq, drv_id); /* fake a irq request */
-	metal_irq_restore_enable(irq_flags_save);
-
-	return 0;
-}
-
 unsigned int metal_irq_save_disable(void)
 {
-	return enter_critical_section();
+	return up_irq_save();
 }
 
 void metal_irq_restore_enable(unsigned int flags)
 {
-	leave_critical_section(flags);
+	up_irq_restore(flags);
 }
 
-void metal_irq_enable(unsigned int vector)
+/* Implement the default irq controller */
+static void metal_cntr_irq_set_enable(struct metal_irq_controller *cntr,
+				      int irq, unsigned int enable)
 {
-	up_enable_irq(vector);
+	if (irq >= 0 && irq < cntr->irq_num) {
+		if (enable == METAL_IRQ_ENABLE)
+			up_enable_irq(irq);
+		else
+			up_disable_irq(irq);
+	}
 }
 
-void metal_irq_disable(unsigned int vector)
+static int metal_cntr_irq_handler(int irq, void *context, void *data)
 {
-	up_disable_irq(vector);
-}
+	if (context != NULL)
+		return metal_irq_handle(data, irq);
 
-/**
-  * @brief       IRQ handler
- */
-static int metal_irq_isr(int irq, void *context, void *arg)
-{
-	struct metal_irq_hddesc *desc = arg;
-
-	if (context != desc->drv_id)
-		return desc->hd(irq, desc->drv_id);
-
-	/* context == drv_id mean unregister */
-	irqchain_detach(irq, metal_irq_isr, arg);
-	sched_kfree(arg);
+	/* context == NULL mean unregister */
+	irqchain_detach(irq, metal_cntr_irq_handler, data);
+	sched_kfree(data);
 	return 0;
+}
+
+static int metal_cntr_irq_attach(struct metal_irq_controller *cntr,
+				 int irq, metal_irq_handler hd, void *arg)
+{
+	if (irq < 0 || irq >= cntr->irq_num)
+		return -EINVAL;
+
+	if (hd) {
+		struct metal_irq *data;
+
+		data = metal_allocate_memory(sizeof(*data));
+		if (data == NULL)
+			return -ENOMEM;
+
+		data->hd  = hd;
+		data->arg = arg;
+
+		irq_attach(irq, metal_cntr_irq_handler, data);
+	} else {
+		unsigned int flags;
+
+		flags = metal_irq_save_disable();
+		irq_dispatch(irq, NULL); /* fake a irq request */
+		metal_irq_restore_enable(flags);
+	}
+
+	return 0;
+}
+
+int metal_cntr_irq_init(void)
+{
+	static METAL_IRQ_CONTROLLER_DECLARE(metal_cntr_irq,
+					    0, NR_IRQS,
+					    NULL,
+					    metal_cntr_irq_set_enable,
+					    metal_cntr_irq_attach,
+					    NULL)
+	return metal_irq_register_controller(&metal_cntr_irq);
 }
--- a/lib/system/nuttx/irq.h
+++ b/lib/system/nuttx/irq.h
@@ -20,6 +20,8 @@
 extern "C" {
 #endif
 
+int metal_cntr_irq_init(void);
+
 #ifdef __cplusplus
 }
 #endif


arnop2:
Hi Wendy,
Please could you remind me why you need to define an IRQ API in libmetal, as this only used by application, not by the OpenAMP lib? 
From my pov, APIs not used by the openAMP library should be implemented in a specific file in the exemple directory, not in the libmetal itself...

lib/irq.h
```diff
@@ -65,6 +103,16 @@ void metal_irq_unregister(int irq)
...
 
+/**
+ * @brief	metal_irq_get_softirq
+ *
+ * Get soft IRQ from triggered from source IRQ
+ *
+ * @param[in] irq source IRQ
+ * @return soft IRQ ID for success, negative value for failure
+ */
+int metal_irq_get_softirq(int irq);
```
xlnxfburton: I don't think irq.[ch] should have any knowledge about soft irq.  Anything related to soft irq should be in softirq.[ch].

lib/irq.h
```diff
@@ -19,21 +19,59 @@ extern "C" {
...
 
+/** IRQ ANY ID */
+#define METAL_IRQ_ANY         (-1)
+
+/** IRQ state macro which will be passed to metal irq set state function
+ *  to indicate which state the caller want the IRQ to change to.
+ */
+#define METAL_IRQ_ENABLE     1U
+#define METAL_IRQ_DISABLE    2U
+#define METAL_IRQ_PENDING    4U
```
xlnxfburton: Making an IRQ pending seems to be specific to soft irq.  could have a function in softirq.[ch] instead of this?  I don't think we need to use the callback for this since it is only applicable to softirq.

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+ *
+ * Map soft IRQ range with input IRQ range which will trigger the soft
+ * IRQs.
+ *
+ * @param[in] source_irq_base Base of source IRQ range.
+ * @param[in] irq_num  Number of IRQs required
+ * @return base of target soft IRQs for success, or negative value for
+ *         failure.
+ */
+int metal_irq_map_softirq(int source_irq_base, int irq_num);
```
xlnxfburton: I don't think we should have a mapping from HW irq to soft irq since there is no guarantee that there is a 1:1 mapping.  often there is no mapping and sometimes there is a 1 to many mapping. I think this should be handled by the user.

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+ *
+ * Allocate IRQ range
+ *
+ * @param[in] irq_base Base of required IRQ range, it can be METAL_IRQ_ANY.
+ *                     If it is METAL_IRQ_ANY, libmetal will allocate the
+ *                     base or IRQ range.
+ * @param[in] irq_num  Number of IRQs required
+ * @return irq number base for success, or negative value for failure.
+ */
+int metal_irq_allocate(int irq_base, int irq_num);
```
xlnxfburton: Do we need a separate function for this?  I would prefer this to be done as part of the metal_irq_register_controller().  This will reduce the number of APIs and I believe metal_irq_register_controller() needs to check that there is no overlap in IRQ ranges anyway.

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+ * @brief	metal_irq_controller_set_range
+ *
+ * Set IRQ controller IRQ range
+ *
+ * @param[in] cntr Interrupt controller
+ * @param[in] irq_base Base IRQ id of the range managed by the controller
+ * @param[in] irq_num  number of IRQs managed by the controller
+ */
+static inline
+void metal_irq_controller_set_range(struct metal_irq_controller *cntr,
```
xlnxfburton: This function is not necessary if we remove metal_irq_allocate() and allocate the range as part of the registration.

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+/**
+ * @brief	metal_irq_controller_set_irqs
+ *
+ * Set IRQ controller IRQs array
+ *
+ * @param[in] cntr Interrupt controller
+ * @param[in] irqs IRQs array
+ */
+static inline
+void metal_irq_controller_set_irqs(struct metal_irq_controller *cntr,
```
xlnxfburton: This function is not necessary since this is initialized using METAL_IRQ_CONTROLLER_DECLARE()

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+ * @brief	metal_irq_set_handler
+ *
+ * Set IRQ handler
+ *
+ * @param[in] irq_data interrupt structure for the handler
+ * @param[in] hd interrupt handler function
+ * @param[in] arg interrupt handler argument
+ */
+static inline
+void metal_irq_set_handler(struct metal_irq *irq_data,
```
xlnxfburton: This function is not used anywhere and it not a public API so I think we should remove it

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+ * @brief	metal_irq_handle
+ *
+ * Call registered IRQ handler
+ *
+ * @param[in] irq_data metal IRQ structure
+ * @param[in] irq IRQ id which will be passed to handler
+ * @return IRQ handler status
+ */
+static inline
+int metal_irq_handle(struct metal_irq *irq_data, int irq)
```
xlnxfburton: This is a helper function for implementing interrupt controllers.  It seems just as easy for interrupt controller implementation to do this directly.  I suggest we remove it from the public API.

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+/**
+ * @brief	metal_irq_controller_set_pending
+ *
+ * Set IRQ to pending state with the IRQ controller
+ *
+ * @param[in] irq IRQ id
+ * @param[in] cntr IRQ controller
+ */
+static inline
+void metal_irq_controller_set_pending(int irq,
```
xlnxfburton: I think this should be moved to softirq.[ch] since it only is applicable to softirq.  This way it does not need to use the call back and HW based interrupt controllers do not need to check for METAL_IRQ_PENDING and return error.

lib/irq.h
```diff
@@ -95,6 +143,158 @@ void metal_irq_enable(unsigned int vector);
...
+}
+
+/**
+ * @brief	metal_irq_set_pending
+ *
+ * Set IRQ to pending state
+ *
+ * @param[in] irq IRQ id
+ */
+void metal_irq_set_pending(int irq);
```
xlnxfburton: I think this should be moved to softirq.[ch] since it only is applicable to softirq. This way it does not need to use the call back and HW based interrupt controllers do not need to check for METAL_IRQ_PENDING and return error.

lib/irq.h
```diff
@@ -19,21 +19,59 @@ extern "C" {
...
+#define METAL_IRQ_ANY         (-1)
+
+/** IRQ state macro which will be passed to metal irq set state function
+ *  to indicate which state the caller want the IRQ to change to.
+ */
+#define METAL_IRQ_ENABLE     1U
+#define METAL_IRQ_DISABLE    2U
+#define METAL_IRQ_PENDING    4U
+
+struct metal_device;
```
xlnxfburton: This is not used by irq.  can we remove it?

wjliang:
> Hi Wendy,
> Please could you remind me why you need to define an IRQ API in libmetal, as this only used by application, not by the OpenAMP lib?
> From my pov, APIs not used by the openAMP library should be implemented in a specific file in the exemple directory, not in the libmetal itself...

Hi Arnaud,

The libmetal IRQ abstraction is to enable userspace or standalone drivers/application to have interrupt handling independent to how interrupt is triggered. We have drivers cross different platforms such as FreeRTOS, baremetal and Linux userspace. The drivers need to register for interrupt handling, for baremetal / RTOS environment, the interrupt is triggered from hardware, for Linux userspace, the interrupt is monitored with file descriptors exposed from kernel device driver such as UIO, VFIO. The driver/application just care about to run something from this notification (IRQ) but don't care about how this notification (IRQ) is triggered. The IRQ abstraction is for this purpose.

Best Regards,
Wendy

lib/irq.c
```diff
@@ -0,0 +1,136 @@
...
+		offset = irq - source_irq_base;
+
+		if (irq >= source_irq_base && offset < irq_maps[i].irq_num) {
+			return irq_maps[i].target_irq_base + offset;
+		}
+	}
+	return -EINVAL;
+}
+
+struct metal_irq_controller *metal_irq_get_controller(int irq)
```
xlnxfburton: This function does not lock the list from changing while it is iterating it.  This could cause the list to be corrupted if we allow multiple threads to dynamically add interrupt controllers.  I cannot think of any use case where we need to add interrupt controllers dynamically or by multiple threads.  Is there a use case for this?  If there is no such use case I suggest to remove the locking from metal_irq_register_controller().

If there are many interrupt controllers then using a list can be inefficient, so we might want to change the list to an array at some point...

lib/irq.c
```diff
@@ -0,0 +1,136 @@
...
+		metal_bitmap_set_bit(metal_irq_allocated, i);
+	}
+	ret = irq_base;
+
+err:
+	atomic_flag_clear(&flag);
+	return ret;
+}
+
+void metal_irq_register_controller(struct metal_irq_controller *cntr)
```
xlnxfburton: I think we should remove the locking from this function and require that all interrupt controllers are registered before multiple threads are used.  Otherwise we would have to do locking in metal_irq_get_controller() as well.

lib/irq.c
```diff
@@ -0,0 +1,136 @@
...
+struct metal_irq_map {
+	int source_irq_base; /**< Source IRQ which triggers the soft IRQ */
+	int target_irq_base; /**< target soft IRQ */
+	int irq_num; /**< number of IRQs of the mapping */
+};
+
+/** Indicate which IRQ has been allocated */
+#define METAL_IRQ_MAX 2048
+static unsigned long
+metal_irq_allocated[metal_div_round_up(METAL_IRQ_MAX, METAL_BITS_PER_ULONG)];
```
xlnxfburton: This will not be needed when metal_irq_allocate() functionality is combined with metal_irq_register_controller()

lib/irq.c
```diff
@@ -0,0 +1,136 @@
...
+
+/** List of registered IRQ controller */
+static METAL_DECLARE_LIST(irq_cntrs);
+
+#define METAL_IRQ_MAX_MAPPINGS 1
+static struct metal_irq_map irq_maps[METAL_IRQ_MAX_MAPPINGS];
+
+int metal_irq_allocate(int irq_base, int irq_num)
```
xlnxfburton: This will not be needed when metal_irq_allocate() functionality is combined with metal_irq_register_controller()

lib/softirq.c
```diff
@@ -0,0 +1,136 @@
...
+	irq_base = metal_irq_allocate(METAL_IRQ_ANY, irq_num);
+	if (irq_base < 0)
+		return -EINVAL;
+	metal_irq_controller_set_range(&metal_softirq_cntr,
+				       irq_base, irq_num);
+	metal_irq_controller_set_irqs(&metal_softirq_cntr, metal_softirqs);
+	return 0;
+}
+
+int metal_softirq_allocate(int num)
```
xlnxfburton: I think we should require softirq allocation to be done before multithreading is started.  This way we don't have to do any locking.  

Do we need a bitmap to keep track of allocated softirq numbers?  Isn't it enough to just have a counter that is incremented for every allocation?

lib/softirq.c
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	for (i = bit; i < bit + num; i++) {
+		metal_bitmap_set_bit(metal_softirq_assigned, i);
+	}
+	atomic_flag_clear(&metal_softirq_lock);
+	return bit;
+}
+
+void metal_softirq_dispatch()
```
xlnxfburton: I think we should require that this function is only called from one thread, for example the main thread.  This way we don't need any locking.

lib/softirq.h
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @brief	metal_softirq_allocate
+ *
+ * Allocate soft IRQs
+ *
+ * @param[in]  num number of soft irqs requested
+ * @return soft irq base for success, or negative value for failure
+ */
+int metal_softirq_allocate(int num);
+
```
xlnxfburton: We need some API to allow sleeping until an interrupt occurs using WFI or similar instructions.

arnop2:
> > Hi Wendy,
> > Please could you remind me why you need to define an IRQ API in libmetal, as this only used by application, not by the OpenAMP lib?
> > From my pov, APIs not used by the openAMP library should be implemented in a specific file in the exemple directory, not in the libmetal itself...
> 
> Hi Arnaud,
> 
> The libmetal IRQ abstraction is to enable userspace or standalone drivers/application to have interrupt handling independent to how interrupt is triggered. We have drivers cross different platforms such as FreeRTOS, baremetal and Linux userspace. The drivers need to register for interrupt handling, for baremetal / RTOS environment, the interrupt is triggered from hardware, for Linux userspace, the interrupt is monitored with file descriptors exposed from kernel device driver such as UIO, VFIO. The driver/application just care about to run something from this notification (IRQ) but don't care about how this notification (IRQ) is triggered. The IRQ abstraction is for this purpose.
> 
> Best Regards,
> Wendy
Thanks for the clarification. For me the libmetal is the abstraction layer for the OpenAMP library, but your definition means that is is more an independent abstraction layer for application and libraries.

lib/irq.h
wjliang: Removed from irq.[ch], after a double thought, remove it from softirq.[ch] as there is already softIRQ api to allocate soft IRQs. Each driver or IRQ controller should be responsible to maintain  the mapping.

lib/irq.h
wjliang: Will remove it from irq.h and will have a softirq function to set the soft IRQ to pending.

lib/irq.h
wjliang: Agree, will remove this from irq.h

lib/irq.h
wjliang: Agree, remove this  function, and let the register to allocate IRQ if required.

lib/irq.h
wjliang: This is just to give the flexibility in case some IRQ controller need to set the IRQ arrays at runtime.
Can remove for now.

lib/irq.h
wjliang: The register IRQ should do the work, will remove.

lib/irq.h
wjliang: Just we have a irq structure to keep the handler and arguments, just think it would be better to wrap around this.

lib/irq.h
wjliang: Agree, will have a softIRQ function to set soft IRQ.

lib/irq.h
wjliang: Will remove the metal_device

wjliang:
> Thanks for the clarification. For me the libmetal is the abstraction layer for the OpenAMP library, but your definition means that is is more an independent abstraction layer for application and libraries.


As an independent library, libmetal library is a thin across OS/platforms abstraction which can be used independent to OpenAMP library.

lib/irq.h
wjliang: Agree, will move this function to softirq.h only.

lib/irq.h
```diff
@@ -19,21 +19,59 @@ extern "C" {
...
+	metal_irq_handler hd; /**< Interrupt handler */
+	void *arg; /**< Argument to pass to the interrupt handler */
+};
+
+/** Libmetal interrupt controller structure */
+struct metal_irq_controller {
+	int irq_base; /**< Start of IRQ number of the range managed by
+		           the IRQ controller */
+	int irq_num; /**< Number of IRQs managed by the IRQ controller */
+	void *arg; /**< Argument to pass to the interrupt handler */
```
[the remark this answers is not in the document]
wjliang: Change it to "Argument to pass to interrupt controller function"

lib/irq.c
wjliang: Will remove metal_irq_get_controller() function from user API. As it is only used internally to call interrupt controller to enable/disable interrupt. It was also used to set the softirq to pending, but this will be covered by a new softirq function. About the locking, agree that we can limit to say that the controller needs to be registered at the initialization time. And not supposed to register controllers by multiple threads. And then remove the locking.

lib/irq.c
wjliang: Agree.

lib/irq.c
wjliang: Yes, will removed the array. And will check the list of the registered controllers to find out the available IRQs.

lib/softirq.c
wjliang: Right, as unlike hard IRQ, there is no predefined soft IRQs, so there should be no holes between the allocated soft IRQs, can remove the bitmap.

lib/softirq.c
```diff
@@ -0,0 +1,136 @@
...
+
+static METAL_IRQ_CONTROLLER_DECLARE(metal_softirq_cntr,
+				    METAL_IRQ_ANY, 0,
+				    NULL,
+				    metal_softirq_set_state,
+				    NULL)
+
+static atomic_flag metal_softirq_lock;
+
+static void metal_softirq_set_enable(int bit, unsigned int state)
```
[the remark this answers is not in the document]
wjliang: We can use atomic flag variable which has a default implementation from libmetal if there is  not implementation in the toolchain. one flag variable per interrupt enables to use atomic operation to set the flag.

lib/softirq.c
```diff
@@ -0,0 +1,136 @@
...
+	if (state ==  METAL_IRQ_ENABLE) {
+		metal_bitmap_set_bit(metal_softirq_enabled, bit);
+	} else {
+		metal_bitmap_clear_bit(metal_softirq_enabled, bit);
+	}
+	atomic_flag_clear(&metal_softirq_lock);
+
+}
+
+static void metal_softirq_set_pending(int bit)
```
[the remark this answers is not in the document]
wjliang: The same as the above comments to the enable flag.

lib/softirq.c
wjliang: I think we don't use bitmap, we don't even need to have this limitation. Even though it should only be called from one thread.

lib/softirq.h
wjliang: why in the softirq allocation function?

lib/irq.c
```diff
@@ -0,0 +1,134 @@
...
+
+	metal_list_for_each(&irq_cntrs, node) {
+		int cntr_irq_base, cntr_irq_end;
+
+		cntr = metal_container_of(node,
+					  struct metal_irq_controller, node);
+		cntr_irq_base = cntr->irq_base;
+		cntr_irq_end = cntr_irq_base + cntr->irq_num;
+		if (irq_tocheck < cntr_irq_end &&
+		    irq_end_tocheck > cntr_irq_base) {
```
xiaoxiang781216: irq_end_tocheck >= cntr_irq_base?

lib/irq.c
```diff
@@ -0,0 +1,134 @@
...
+
+	metal_list_for_each(&irq_cntrs, node) {
+		int irq_base, irq_end;
+
+		cntr = (struct metal_irq_controller *)
+		       metal_container_of(node, struct metal_irq_controller,
+				          node);
+		irq_base = cntr->irq_base;
+		irq_end = irq_base + cntr->irq_num;
+		if (irq >= irq_base && irq <= irq_end) {
```
xiaoxiang781216: irq < irq_end

lib/softirq.h
xiaoxiang781216: > We need some API to allow sleeping until an interrupt occurs using WFI or similar instructions.

Need abstraction here, don't call wfi directly. In RTOS context, one thread idle doesn't mean the whole system is idle.

lib/system/freertos/xlnx_common/irq.c
```diff
@@ -0,0 +1,73 @@
...
+{
+	struct metal_irq *irq_data;
+
+	if (vector > MAX_IRQS) {
+		return;
+	}
+	irq_data = &irqs[vector];
+	if (irq_data->hd) {
+		(void)irq_data->hd(vector, irq_data->arg);
+	}
```
xiaoxiang781216: reuse metal_irq_handle here

lib/system/generic/xlnx_common/irq.c
```diff
@@ -0,0 +1,73 @@
...
+
+#define MAX_IRQS XLNX_MAXIRQS
+
+static struct metal_irq irqs[MAX_IRQS]; /**< Linux IRQs array */
+
+static void metal_xlnx_irq_set_enable(struct metal_irq_controller *irq_cntr,
+				      int irq, unsigned int state)
+{
+	if (irq < irq_cntr->irq_base ||
+	    irq > irq_cntr->irq_base + irq_cntr->irq_num) {
```
xiaoxiang781216: irq >= irq_cntr->irq_base + irq_cntr->irq_num

lib/system/generic/xlnx_common/irq.c
```diff
@@ -0,0 +1,73 @@
...
+				    irqs)
+
+/**
+ * @brief default handler
+ */
+void metal_xlnx_irq_isr(unsigned int vector)
+{
+	struct metal_irq *irq_data;
+
+	if (vector > MAX_IRQS) {
```
xiaoxiang781216: vector >= MAX_IRQS

lib/system/generic/xlnx_common/irq.c
```diff
@@ -0,0 +1,73 @@
...
+		return;
+	}
+	irq_data = &irqs[vector];
+	if (irq_data->hd) {
+		(void)irq_data->hd(vector, irq_data->arg);
```
xiaoxiang781216: reuse metal_irq_handle here

lib/system/freertos/xlnx_common/irq.c
```diff
@@ -0,0 +1,73 @@
...
+				    irqs)
+
+/**
+ * @brief default handler
+ */
+void metal_xlnx_irq_isr(unsigned int vector)
+{
+	struct metal_irq *irq_data;
+
+	if (vector > MAX_IRQS) {
```
xiaoxiang781216: vector >= MAX_IRQS

lib/system/freertos/xlnx_common/irq.c
```diff
@@ -0,0 +1,73 @@
...
+
+#define MAX_IRQS XLNX_MAXIRQS
+
+static struct metal_irq irqs[MAX_IRQS]; /**< Linux IRQs array */
+
+static void metal_xlnx_irq_set_enable(struct metal_irq_controller *irq_cntr,
+				     int irq, unsigned int state)
+{
+	if (irq < irq_cntr->irq_base ||
+	    irq > irq_cntr->irq_base + irq_cntr->irq_num) {
```
xiaoxiang781216: > to >=

lib/irq.h
```diff
@@ -19,61 +19,85 @@ extern "C" {
...
+
+/** Libmetal interrupt controller structure */
+struct metal_irq_controller {
+	int irq_base; /**< Start of IRQ number of the range managed by
+		           the IRQ controller */
+	int irq_num; /**< Number of IRQs managed by the IRQ controller */
+	void *arg; /**< Argument to pass to interrupt controller function */
+	metal_irq_set_enable irq_set_enable; /**< function to set IRQ eanble */
+	struct metal_list node; /**< list node */
+	struct metal_irq *irqs; /**< Array of IRQs managed by the controller */
```
xiaoxiang781216: Since RTOS with irq abstraction already have the similar irqs array, could you add metal_irq_register callback? so we can overwrite the callback and pass the information directly to OS API.

lib/irq.h
```diff
@@ -19,61 +19,85 @@ extern "C" {
...
+				     int irq, unsigned int enable);
+
+/** Libmetal interrupt structure */
+struct metal_irq {
+	metal_irq_handler hd; /**< Interrupt handler */
+	void *arg; /**< Argument to pass to the interrupt handler */
+};
+
+/** Libmetal interrupt controller structure */
+struct metal_irq_controller {
```
xiaoxiang781216: It's better to split metal_irq_controller and related API to another header file(e.g. irqchip.h or irqcontroller.h).
1.metal_irq_register/metal_irq_enable is for application
2.metal_irq_register_controller is for porting layer

lib/irq.h
xiaoxiang781216: Or provide a easy method to bypass the default metal_irq(base on metal_irq_controller)  impelementation

lib/softirq.c
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	atomic_store(&metal_softirq_pending[irq], 1);
+}
+
+int metal_softirq_init()
+{
+	int ret;
+
+	ret = metal_irq_register_controller(&metal_softirq_cntr);
```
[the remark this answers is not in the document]
wjliang: ok, will remove it but return the metal_irq_register_controller() instead

lib/softirq.c
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void metal_softirq_dispatch()
+{
+	int i;
+
+	for (i = 0; i < metal_softirq_num; i++) {
+		struct metal_irq *irq;
+
+		if (metal_softirq_pending[i] != 0 &&
```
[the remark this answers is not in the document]
wjliang: Agree, will use atomic_compare_exchange() to check and if irq is pending.

lib/system/freertos/xlnx_common/irq.c
wjliang: I don't really get this comment, do  you mean prefer to use a metal_irq_handler to get the hd first?
Will the compiler do the optimisation automatically?

lib/irq.c
wjliang: it will be ">" as the "irq_end_tocheck" is "irq_tocheck + irq_num" but not "irq_tocheck + irq_num - 1"

lib/irq.c
wjliang: ok, will make the change.

lib/system/generic/xlnx_common/irq.c
wjliang: will make the change.

lib/system/freertos/xlnx_common/irq.c
xiaoxiang781216: we can simplify to one line code:
metal_irq_handle(irqs[vector], vector);

lib/irq.c
xiaoxiang781216: Emm, you are right.

lib/system/generic/xlnx_common/irq.c
wjliang: I don't get the comment. do you suggest to have a metal_irq_handle variable?

lib/system/freertos/xlnx_common/irq.c
wjliang: will change

lib/system/freertos/xlnx_common/irq.c
wjliang: will change

lib/irq.h
wjliang: Ok, will have metal_irq_register callback of the controller struct.

wjliang:
> > Hi Wendy,
> > Please could you remind me why you need to define an IRQ API in libmetal, as this only used by application, not by the OpenAMP lib?
> > From my pov, APIs not used by the openAMP library should be implemented in a specific file in the exemple directory, not in the libmetal itself...
> 
> Hi Arnaud,
> 
> The libmetal IRQ abstraction is to enable userspace or standalone drivers/application to have interrupt handling independent to how interrupt is triggered. We have drivers cross different platforms such as FreeRTOS, baremetal and Linux userspace. The drivers need to register for interrupt handling, for baremetal / RTOS environment, the interrupt is triggered from hardware, for Linux userspace, the interrupt is monitored with file descriptors exposed from kernel device driver such as UIO, VFIO. The driver/application just care about to run something from this notification (IRQ) but don't care about how this notification (IRQ) is triggered. The IRQ abstraction is for this purpose.
> 
> Best Regards,
> Wendy

Also, the existing irq API in libmetal doesn't have abstraction for irq controller drivers, this patch is to fix that.

wjliang:
@xlnxfburton, @xiaoxiang-xiaomi, @xiaoxiang781216, I have updated the patch based on your comments. Please take a look.

lib/softirq.c
```diff
@@ -0,0 +1,97 @@
...
+{
+	int i;
+
+	for (i = 0; i < metal_softirq_num; i++) {
+		struct metal_irq *irq;
+		char is_pending = 1;
+
+		if (metal_softirq_enabled[i] != 0 &&
+		    atomic_compare_exchange_strong(&metal_softirq_pending[i],
+						   &is_pending, 1)) {
```
xiaoxiang781216: change to:
atomic_compare_exchange_strong(&metal_softirq_pending[i],
							   &is_pending, 0))

wjliang: Changed.

lib/irq.c
```diff
@@ -0,0 +1,138 @@
...
+{
+	struct metal_irq_controller *cntr;
+	struct metal_irq *irq_data;
+
+	cntr = metal_irq_get_controller(irq);
+	if (cntr == NULL) {
+		return -EINVAL;
+	}
+	if (cntr->irq_register != NULL) {
+		return cntr->irq_register(cntr, irq, irq_handler);
```
xiaoxiang781216: need pass arg to callback too

lib/softirq.c
```diff
@@ -0,0 +1,97 @@
...
+static void metal_softirq_set_enable(struct metal_irq_controller *cntr,
+				     int irq, unsigned int enable)
+{
+	if (irq < cntr->irq_base ||
+	    irq >= (cntr->irq_base + cntr->irq_num)) {
+		return;
+	}
+
+	if (enable ==  METAL_IRQ_ENABLE) {
+		atomic_store(&metal_softirq_enabled[irq], 1);
```
xiaoxiang781216: irq need sub cntr->irq_base

lib/softirq.c
```diff
@@ -0,0 +1,97 @@
...
+	struct metal_irq_controller *cntr;
+
+	cntr = &metal_softirq_cntr;
+
+	if (irq < cntr->irq_base ||
+	    irq >= (cntr->irq_base + cntr->irq_num)) {
+		return;
+	}
+
+	atomic_store(&metal_softirq_pending[irq], 1);
```
xiaoxiang781216: irq need sub cntr->irq_base

lib/softirq.c
```diff
@@ -0,0 +1,97 @@
...
+{
+	if (irq < cntr->irq_base ||
+	    irq >= (cntr->irq_base + cntr->irq_num)) {
+		return;
+	}
+
+	if (enable ==  METAL_IRQ_ENABLE) {
+		atomic_store(&metal_softirq_enabled[irq], 1);
+	} else {
+		atomic_store(&metal_softirq_enabled[irq], 0);
```
xiaoxiang781216: irq need sub cntr->irq_base

lib/softirq.c
```diff
@@ -0,0 +1,97 @@
...
+
+	for (i = 0; i < metal_softirq_num; i++) {
+		struct metal_irq *irq;
+		char is_pending = 1;
+
+		if (metal_softirq_enabled[i] != 0 &&
+		    atomic_compare_exchange_strong(&metal_softirq_pending[i],
+						   &is_pending, 0)) {
+			irq = &metal_softirqs[i];
+			(void)metal_irq_handle(irq, i);
```
xiaoxiang781216: i need add cntr->irq_base

lib/irq.c
wjliang: Will change

lib/softirq.c
wjliang: Will change

lib/softirq.c
wjliang: will change

lib/softirq.c
wjliang: will change

lib/softirq.c
wjliang: will change

wjliang:
@xiaoxiang781216, updated patches based on your comments, please check. Thanks, Wendy

xiaoxiang781216:
> @xiaoxiang781216, updated patches based on your comments, please check. Thanks, Wendy
The change looks good for me now.
I also change nuttx porting to utilize the new irq controller interface, please review the attachment:
[0001-nuttx-implement-the-default-irq-controller.zip](https://github.com/OpenAMP/libmetal/files/2816234/0001-nuttx-implement-the-default-irq-controller.zip)

PR #73, opened 2019-01-04
compiler.h: add METAL_PACKED_BEGIN and METAL_PACKED_END

xiaoxiang781216:
Change-Id: I9d87ca5ea95ced0d81c186fb08adc81c69755ce9


xiaoxiang781216:
@wjliang, this patch add macro to define the packed struct, the next patch will remove the similar macro in openamp.

PR #80, opened 2019-02-25
nuttx: handle the no zero virtual address correctly

xiaoxiang781216:
Change-Id: I809472bfecfe18179d4a3ace47c26fe0baa2e771

### DIFF
--- a/lib/system/nuttx/io.c
+++ b/lib/system/nuttx/io.c
@@ -75,13 +75,13 @@ static void metal_io_close_(struct metal_io_region *io)
 static metal_phys_addr_t metal_io_offset_to_phys_(struct metal_io_region *io,
 						  unsigned long offset)
 {
-	return up_addrenv_va_to_pa((void *)offset);
+	return up_addrenv_va_to_pa((char *)io->virt + offset);
 }
 
 static unsigned long metal_io_phys_to_offset_(struct metal_io_region *io,
 					      metal_phys_addr_t phys)
 {
-	return (unsigned long)up_addrenv_pa_to_va(phys);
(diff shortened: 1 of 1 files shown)


lib/system/nuttx/io.c
```diff
@@ -75,13 +75,13 @@ static void metal_io_close_(struct metal_io_region *io)
 static metal_phys_addr_t metal_io_offset_to_phys_(struct metal_io_region *io,
 						  unsigned long offset)
 {
-	return up_addrenv_va_to_pa((void *)offset);
+	return up_addrenv_va_to_pa(io->virt + offset);
```
wjliang: pointer void * + offset, this will not work for all compilers. As the size of *(void * p) is not determined.
Better to use char *

lib/system/nuttx/io.c
```diff
@@ -75,13 +75,13 @@ static void metal_io_close_(struct metal_io_region *io)
...
 }
 
 static unsigned long metal_io_phys_to_offset_(struct metal_io_region *io,
 					      metal_phys_addr_t phys)
 {
-	return (unsigned long)up_addrenv_pa_to_va(phys);
+	return up_addrenv_pa_to_va(phys) - io->virt;
```
wjliang: don't you need casting?

xiaoxiang781216: Fixed.

lib/system/nuttx/io.c
xiaoxiang781216: Fixed

xiaoxiang781216:
@wjliang, the new patch fix the cast issue, please review.

Issue #78, opened 2019-02-06
__builtin_assume_aligned() is gnu gcc only


wjliang:
__builtin_assume_aligned() is gnu gcc only, need to have a macro for it.

arnopo:
close as __builtin_assume_aligned() no more used

PR #81, opened 2019-02-25
Revert "utilities: restrict the bitmap pointer is aligned"

xiaoxiang781216:
since no all compiler support __builtin_assume_aligned
This reverts commit 0b979d4cbc9c51044664340539d66862597f3172.

Change-Id: I7c3523b63331917f3a330e08ba4196e1f1534daf

### DIFF
--- a/lib/utilities.h
+++ b/lib/utilities.h
@@ -78,7 +78,6 @@ extern "C" {
 
 static inline void metal_bitmap_set_bit(unsigned long *bitmap, int bit)
 {
-	bitmap = __builtin_assume_aligned(bitmap, sizeof(unsigned long));
 	bitmap[bit / METAL_BITS_PER_ULONG] |=
 		metal_bit(bit & (METAL_BITS_PER_ULONG - 1));
 }
@@ -91,7 +90,6 @@ static inline int metal_bitmap_is_bit_set(unsigned long *bitmap, int bit)
 
 static inline void metal_bitmap_clear_bit(unsigned long *bitmap, int bit)
 {
-	bitmap = __builtin_assume_aligned(bitmap, sizeof(unsigned long));
 	bitmap[bit / METAL_BITS_PER_ULONG] &=
 		~metal_bit(bit & (METAL_BITS_PER_ULONG - 1));
 }


lib/utilities.h
```diff
@@ -78,7 +78,6 @@ extern "C" {
 
 static inline void metal_bitmap_set_bit(unsigned long *bitmap, int bit)
 {
-	bitmap = __builtin_assume_aligned(bitmap, sizeof(unsigned long));
```
wjliang: better to defined METAL_ASSUME_ALIGNED macro.
But this function also has issue, will also need to restrict the input bitmap so that the compiler can do optimization.

wjliang:
This is related to issue: https://github.com/OpenAMP/libmetal/issues/78

lib/utilities.h
xiaoxiang781216: Two questions here:
1.bitmap is "unsigned long" type and should already be aligned to sizeof(unsigned long), why we need __builtin_assume_aligned again?
2.metal_bitmap_set_bit/metal_bitmap_clr_bit is only used by metal_linux_irq_set_enable, I suppose the calling frequency isn't very high. Why we need optimize the corner case here?

wjliang: > Two questions here:
> 1.bitmap is "unsigned long" type and should already be aligned to sizeof(unsigned long), why we need __builtin_assume_aligned again?
Just the input is (unsigned long *) i cannot see limitation on casting a not unsigned long aligned address to (unsigned long *) pointer.
> 2.metal_bitmap_set_bit/metal_bitmap_clr_bit is only used by metal_linux_irq_set_enable, I suppose the calling frequency isn't very high. Why we need optimize the corner case here?
the metal_bitmap_set/clr_bit can be used directly by the application/library on top of libmetal, it is not an internal API.
my point to this patch is instead of removing it, it is good to have a macro to define the assume_aligned.

xiaoxiang781216: But as item 1 point out, bitmap is already 4/8 bytes alignment. At least for this particular case, 
__builtin_assume_aligned is redundant and confused. So it's better to revert the change and wait the real case to add the portable macro back.

wjliang: > But as item 1 point out, bitmap is already 4/8 bytes alignment. At least for this particular case,
> __builtin_assume_aligned is redundant and confused. So it's better to revert the change and wait the real case to add the portable macro back.

It was supposed to for the compiler to optimize the function, without it, it will not do the optimisation.
But since it is not enough, it is OK to remove it for now, just prefer to to fix it properly instead of removing it. But I will take this patch.

PR #82, opened 2019-02-26
[RFC] Update libmetal shared memory API

wjliang:
This patch series is to provide a cross OS abstraction to allow allocate shared memory from any memory allocator, and allow to shared the memory with different processes, devices or coprocessors.

To shared memory with device from Linux, the implementation of this patch make use of ION implementation which is in staging directory from Linux kernel. With ION, you can request DMA buffer from a particular ION heap, and then, you can assign the DMA buf to a device. You can sync the shared memory between the CPUs and the devices with the DMA buf operations.

There is no generic DMA buf APIs to export the DMA buf to a device in kernel, this pull request uses the UIO ioctl  from this patch: https://lkml.org/lkml/2019/2/23/207 to export the DMA buf to an UIO device.

It can be extended to any other backend.

v4:
* fix typo "breif" to "brief"

v3:
* Use uintptr_t to cast __u64 to pointer and vice versa

v2:
* update the UIO shm attachment API to also return virtual address.


starhwk:
Hi Wendy, I integrated this and tested with my own local module. Some basic test that allocates buffers and uses it with dma works. Hence,
Tested-by: Hyun Kwon <hyun.kwon@xilinx.com>

edmooring:
Given that this PR depends on a kernel feature that is not yet upstreamed, is there any point in keeping it hanging on?

PR #85, opened 2019-03-01
Fix atomic operations compilation errors

wjliang:
These patches fix the places which do not comply with the c11 atomic operations specification.

### DIFF
--- a/lib/spinlock.h
+++ b/lib/spinlock.h
@@ -22,11 +22,11 @@ extern "C" {
 /** \defgroup spinlock Spinlock Interfaces
  *  @{ */
 struct metal_spinlock {
-	atomic_int v;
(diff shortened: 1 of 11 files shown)


wjliang:
@galak , @arnop2 please review

lib/spinlock.h
```diff
@@ -22,7 +22,7 @@ extern "C" {
 /** \defgroup spinlock Spinlock Interfaces
  *  @{ */
 struct metal_spinlock {
-	atomic_int v;
+	atomic_flag v;
 };
```
galak: METAL_SPINLOCK_INIT needs to change to use ATOMIC_FLAG_INIT(0)

wjliang:
@galak , updated based on your comments

PR #94, opened 2019-05-23
 IAR c-stat reports

arnopo:
This series proposes corrections for few compilation issues and issues reported by IAR c-stat tools (alignment with MISRA C 2012)
As the code is platform dependent , i fixed only the the common part.

### DIFF
--- a/lib/shmem.c
+++ b/lib/shmem.c
@@ -38,10 +38,10 @@ int metal_shmem_open_generic(const char *name, size_t size,
 		shmem = metal_container_of(node, struct metal_generic_shmem, node);
 		if (strcmp(shmem->name, name) != 0)
 			continue;
-		if (size > metal_io_region_size(&shmem->io))
-			continue;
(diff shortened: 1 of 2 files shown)


lib/shmem.c
```diff
@@ -38,10 +38,10 @@ int metal_shmem_open_generic(const char *name, size_t size,
 		shmem = metal_container_of(node, struct metal_generic_shmem, node);
 		if (strcmp(shmem->name, name) != 0)
 			continue;
-		if (size > metal_io_region_size(&shmem->io))
-			continue;
-		*result = &shmem->io;
-		return 0;
+		if (size < metal_io_region_size(&shmem->io)) {
```
edmooring: What about the case where size==metal_io_region_size()? This change makes the code behave differently. Maybe "<="?

arnop2: fixed, thanks!

PR #96, opened 2019-06-24
lib: arm CM33: Add Cortex M33 support

karl-zh:
TFM (Trusted Firmware for M class) is a trusted firmware running on arm M
series core to provided security functions.

This patch aims to add support TFM build libmetal.

Signed-off-by: Karl Zhang <karl.zhang@linaro.org>


karl-zh:
@wjliang Thanks for the responding. I see now.

PR #100, opened 2019-09-18
Fix C++ compilation errors on atomic types

mizmar:
Includes were moved out from of `extern "C"` blocks. This is discouraged practice as the block also applies on included header and was affecting the include of `<atomic>` header in [lib/atomic.h](https://github.com/OpenAMP/libmetal/blob/edb1c31155839de022e33a74b48914b5575a9047/lib/atomic.h#L22) and causing errors.

Following the instructions in [P0943R0: Support C atomics in C++](http://www.open-std.org/JTC1/SC22/WG21/docs/papers/2018/p0943r0.html), imports were added to make types compatible with `<stdatomic.h>`.

[metal_mutex_t](https://github.com/OpenAMP/libmetal/blob/edb1c31155839de022e33a74b48914b5575a9047/lib/system/generic/mutex.h#L26) was probably meant to use `atomic_flag` as it is used with `atomic_flag_*` functions, but according to C++ [documentation](https://en.cppreference.com/w/cpp/atomic/atomic_flag), `atomic_flag` doesn't provide load and store methods and underlying type is supposed to be boolean, thus cast to `atomic_bool*` seemed appropriate. There has to be a better way to solve this.

### DIFF
--- a/lib/alloc.h
+++ b/lib/alloc.h
@@ -31,7 +31,7 @@ static inline void *metal_allocate_memory(unsigned int size);
 /**
  * @brief      free the memory previously allocated
  *
- * @param[in]  ptr       pointer to memory 
+ * @param[in]  ptr       pointer to memory
  */
 static inline void metal_free_memory(void *ptr);
 
--- a/lib/alloc.h
+++ b/lib/alloc.h
@@ -35,12 +35,12 @@ static inline void *metal_allocate_memory(unsigned int size);
  */
 static inline void metal_free_memory(void *ptr);
 
-#include <metal/system/@PROJECT_SYSTEM@/alloc.h>
-
 /** @} */
 
 #ifdef __cplusplus
 }
 #endif
+
+#include <metal/system/@PROJECT_SYSTEM@/alloc.h>
 
 #endif /* __METAL_ALLOC__H__ */
--- a/lib/atomic.h
+++ b/lib/atomic.h
@@ -20,6 +20,15 @@
 # include <stdatomic.h>
 #elif defined(__cplusplus)
 # include <atomic>
+# define _Atomic(T) std::atomic<T>
+using std::memory_order;
+using std::atomic_flag;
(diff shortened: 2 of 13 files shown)


edmooring:
Thanks very much for the patch. Do you have a test program you can share?

mizmar:
I was intending to use OpenAMP (and thus libmetal) as shipped in [STM32CubeMP1 Package](https://github.com/STMicroelectronics/STM32CubeMP1) by STMicroelectronics. This package contains an old release of both projects and ports for STM32MP157xx boards. I haven't try much beside examples provided in this package and its use in C++ project.

I've used following setup and program: [openamp-project.zip](https://github.com/OpenAMP/libmetal/files/3633502/openamp-project.zip), it uses up-to-date libmetal and OpenAMP with patches in question.

Compilation instructions:
- install [GNU Arm Embedded Toolchain](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm) into PATH
- install [Meson](https://mesonbuild.com/) build system
- `meson build --cross cross-stm32mp1.txt`
- `ninja -C build`

arnopo:
@mizmar 
Please could you split your PR in two: 
- one to fix formating and include
- one for the atomic ( this one needs more investigation to ensure non regression)

you also need to follow the same coding rules than other commits for your commit message formatting ( header, sign-off...).

arnopo:
@mizmar 
Do you plan to update your patchset according to remarks, that we can go forward?
Please  notice also that you will have to rebase it.
Thank,

sleiner:
I did another fix for the underlying problem in #126 and #127

arnopo:
Hi @sleiner
What is the next step for this PR? Do you consider that we can close it? or do you plan to update it to propose some potential remaining fixes?

sleiner:
We can close it 👍 #127 essentially does all the same things 😄

PR #104, opened 2019-11-09
nuttx: correctly handle the hardware which hasn't the interrupt controller

xiaoxiang781216:


xiaoxiang781216:
@edmooring please take a look.

xiaoxiang781216:
> Please add text to the commit message explaining why this change was made and what problem it solves. This will make it easier for future readers of the code to understand.

Please take a look.

edmooring:
Signed-off-by: Ed Mooring <ed.mooring@linaro.org>

Issue #107, opened 2020-03-24
CMAKE_BUILD_TYPE is not permitted to be empty string


SebastianBoe:
Having CMAKE_BUILD_TYPE be the empty string is a valid configuration. But options.cmake will change it to Debug when it is empty string, thereby preventing us from using this valid configuration.

https://cmake.org/cmake/help/v3.17/variable/CMAKE_BUILD_TYPE.html

> This statically specifies what build type (configuration) will be built in this build tree. Possible values are empty, Debug, Release, RelWithDebInfo, MinSizeRel, …

PR #108, opened 2020-03-24
cmake: Add support for having CMAKE_BUILD_TYPE be the empty string

SebastianBoe:
Add support for having CMAKE_BUILD_TYPE be the empty string.

This fixes https://github.com/zephyrproject-rtos/zephyr/issues/23713
and
https://github.com/OpenAMP/libmetal/issues/107

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>


cmake/options.cmake
```diff
@@ -3,9 +3,9 @@ set (PROJECT_VER_MINOR  1)
 set (PROJECT_VER_PATCH  0)
 set (PROJECT_VER        0.1.0)
 
-if (NOT CMAKE_BUILD_TYPE)
+if (NOT DEFINED CMAKE_BUILD_TYPE)
   set (CMAKE_BUILD_TYPE Debug)
-endif (NOT CMAKE_BUILD_TYPE)
+endif (NOT DEFINED CMAKE_BUILD_TYPE)
```
arnopo: I think  both condition above CMAKE_INSTALL_LIBDIR && CMAKE_INSTALL_BINDIR should be reworked also. 
anyway let's fix this first

SebastianBoe:
Commit message is updated

PR #113, opened 2020-04-21
nuttx: change sched_kfree to kmm_free

xiaoxiang781216:
since the upstream remove sched_kfree and put the similar logic into kmm_free


lib/system/nuttx/irq.c
```diff
@@ -45,7 +45,7 @@ static int metal_cntr_irq_handler(int irq, void *context, void *data)
 
 	/* context == NULL mean unregister */
 	irqchain_detach(irq, metal_cntr_irq_handler, data);
-	sched_kfree(data);
+	kmm_free(data);
```
arnopo, line 48: As metal_allocate_memory is used you should use metal_free_memory ( which is relying on kmm_free)

xiaoxiang781216: Good finding!

PR #119, opened 2020-06-19
cmake: set the policy CMP0077 to NEW

tejlmand:
This commit introduces CMP0077 and set the policy to new behavior
and thus removes the following warning when using newer CMake versions:
```
Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
--help-policy CMP0077" for policy details.  Use the cmake_policy command
to set the policy and suppress this warning.`
```

Both OLD and NEW behavior will use the value defined earlier in the
CMake process, however the OLD behavior will also add an entry to the
CMakeCache.txt but ignore that value on sub-sequent invocations.
As this may lead to confusion on users looking into the CMakeCache.txt,
then the best is to not get such values into the cache, and thus use NEW
behavior.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,10 @@ if (POLICY CMP0048)
   cmake_policy(SET CMP0048 NEW)
 endif()
 
+if (POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
 list (APPEND CMAKE_MODULE_PATH
       "${CMAKE_CURRENT_SOURCE_DIR}/cmake"
       "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules"


tejlmand:
Seems travis-ci uses an too old Python for Zephyr building:
```
 if [[ "$TARGET" == "zephyr" ]]; then mkdir -p ../libmetal/build-zephyr && cd ../libmetal/build-zephyr && cmake .. -DWITH_ZEPHYR=on -DBOARD=qemu_cortex_m3 && make VERBOSE=1; fi
-- Application: /home/travis/build/OpenAMP/libmetal
-- Zephyr version: 2.3.99 (/home/travis/build/OpenAMP/zephyr)
CMake Error at /home/travis/bin/cmake/cmake-3.13.1-Linux-x86_64/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find Python3: Found unsuitable version "3.4.3", but required is
  at least "3.6" (found /usr/bin/python3)
```
can someone look into travis ?

arnopo:
@tejlmand
i don't have the warning for  libmetal...
Could you confirm that you have it? and in this case provide the cmake command used and the cmake version?

Thanks
Arnaud

arnopo:
> Seems travis-ci uses an too old Python for Zephyr building:
> 
> ```
>  if [[ "$TARGET" == "zephyr" ]]; then mkdir -p ../libmetal/build-zephyr && cd ../libmetal/build-zephyr && cmake .. -DWITH_ZEPHYR=on -DBOARD=qemu_cortex_m3 && make VERBOSE=1; fi
> -- Application: /home/travis/build/OpenAMP/libmetal
> -- Zephyr version: 2.3.99 (/home/travis/build/OpenAMP/zephyr)
> CMake Error at /home/travis/bin/cmake/cmake-3.13.1-Linux-x86_64/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
>   Could NOT find Python3: Found unsuitable version "3.4.3", but required is
>   at least "3.6" (found /usr/bin/python3)
> ```
> 
> can someone look into travis ?

yes we need to migrate the test to a github action. Please don't take into account zephyr result until migration.

tejlmand:
> @tejlmand
> i don't have the warning for libmetal...
> Could you confirm that you have it? and in this case provide the cmake command used and the cmake version?
> 
> Thanks
> Arnaud

I did a little more investigation on this, and the reason you don't see it directly in libmetal, is because the option is not being shadowed.
In Zephyr, the following is done:
https://github.com/zephyrproject-rtos/libmetal/blob/master/CMakeLists.txt#L10-L13

and thus in Zephyr we see the warning:
```
CMake Warning (dev) at /projects/github/ncs/modules/hal/libmetal/libmetal/cmake/options.cmake:55 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
```

before libmetal is included, so that it is ensured the libmetal is built for Zephyr.
For test, you can try to apply:
```
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -40,6 +40,7 @@ string (TOUPPER ${CMAKE_SYSTEM_PROCESSOR} PROJECT_PROCESSOR_UPPER)
 string (TOLOWER ${MACHINE}                PROJECT_MACHINE)
 string (TOUPPER ${MACHINE}                PROJECT_MACHINE_UPPER)
 
+set(WITH_STATIC_LIB 1)
 option (WITH_STATIC_LIB "Build with a static library" ON)
 
 if ("${PROJECT_SYSTEM}" STREQUAL "linux")

```
and you'll see:
```
CMake Warning (dev) at cmake/options.cmake:44 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'WITH_STATIC_LIB'.
```

So actually it might be preferred to set CMP0077 to NEW instead:
https://cmake.org/cmake/help/latest/policy/CMP0077.html

But then the question is, what to do for CMake <3.12 users ?

Let me do a bit more investigation on this.

tejlmand:
DNM to allow some more discussion on CMake version.

tejlmand:
@arnopo I finally found the time to investigate this a bit more.
Both OLD and NEW behavior will behave the same, in the sense that the both will honor the current value of the option, if specified earlier in the build system, such as here: https://github.com/zephyrproject-rtos/open-amp/blob/master/CMakeLists.txt#L9

The difference between OLD and NEW is that OLD will add an entry in the CMakeCache.txt, which is set to either ON / OFF, irregardless of the previous value, so in fact the value from the CMakeCache.txt is effectively ignored.

Using the CMP0077 NEW ensures the variable will never enter the CMakeCache.txt and thus remove any risk of confusing users who examines the CMakeCache.txt.

Ready for merge.

PR #124, opened 2020-07-21
[RFC] zephyr: make libmetal a Zephyr module

b0661:
Allow libmetal to be used as a Zephyr module. This allows easy integration
into Zephyr projects.

OpenAMP libmetal can be used instead of the Zephyr fork. Configuration in
Zephyr is distinct for the two `libmetals`.

Testing has not been adapted - the test knowingly fails.

Making also open-amp a Zephyr module is in my pipeline. I will provide a pull request
to open-amp if this pull request is accepted.

Signed-off-by: Bobby Noelte <b0661n0e17e@gmail.com>

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -64,42 +64,35 @@ if (WITH_DEFAULT_LOGGER)
   add_definitions (-DDEFAULT_LOGGER_ON)
 endif (WITH_DEFAULT_LOGGER)
 
-if (WITH_ZEPHYR)
-  zephyr_library_named(metal)
-  add_dependencies(metal offsets_h)
-  zephyr_library_sources(${_sources})
-  zephyr_include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
-else (WITH_ZEPHYR)
-  # Build a shared library if so configured.
-  if (WITH_SHARED_LIB)
-    set (_lib ${PROJECT_NAME}-shared)
-    add_library (${_lib} SHARED ${_sources})
-    target_link_libraries (${_lib} ${_deps})
-    install (TARGETS ${_lib} LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
-    if (PROJECT_EC_FLAGS)
-      string(REPLACE " " ";" _ec_flgs ${PROJECT_EC_FLAGS})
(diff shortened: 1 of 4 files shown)


arnopo:
Very interesting! would be nice to avoid the fork in Zephyr.
I would suggest that you first propose this first in Zephyr for libmetal and openamp libraries. we will back port it when approved in Zephyr.

edmooring:
I would agree with Arnaud. I'd like to see this in Zephyr first.

b0661:
> I would agree with Arnaud. I'd like to see this in Zephyr first.

Zephyr does not need that. @galak already adapted it to be a Zephyr module and changed the CMakeLists.txt file to fit to that.
The difference here is that the CMakeLists.txt file for the Zephyr module can coexist to the usual one for libmetal but still uses the same underlying infrastructure.

arnopo:
> > I would agree with Arnaud. I'd like to see this in Zephyr first.
> 
> Zephyr does not need that. @galak already adapted it to be a Zephyr module and changed the CMakeLists.txt file to fit to that.
> The difference here is that the CMakeLists.txt file for the Zephyr module can coexist to the usual one for libmetal but still uses the same underlying infrastructure.

It seems to me that we first need to clarify your request...

So the objective is not to avoid the "fork in zephyr" project, is it?

Today, the integration of libmetal and openamp is already based on a module and is quite simple to integrate the lib updates in zephyr (only the cleaning of a few directories). I can not see the improvement introduced here for the integration.
So could you please clarify the purpose of this RFC.

Having said that, removing the "WITH_ZEPHYR" option would be nice. But IMO, the files in the "./zephyr" directory should be part of the Zephyr module, and of the test/system/zephyr directory not at the top of the libmetal tree .  
So if you want to propose an evolution, you should probably do it in 2 steps.
- adapt the test environment of the metal library and the openamp library to match your lib/CMakeLists.txt update proposal
- update the zephyr modules accordingly (by merging your patch linked to the zephyr directory in the zephyr module).

b0661:
> Today, the integration of libmetal and openamp is already based on a module and is quite simple to integrate the lib updates in  zephyr (only the cleaning of a few directories). I can not see the improvement introduced here for the integration

I see your point. Thank you for the review. I will rethink the integration strategy and close this for now.

PR #125, opened 2020-07-22
Compatibility with pure C11 clients

sleiner:
This patch makes a test file which includes all metal headers compile with the following arguments: `-std=c11 -pedantic`.

Tested with GCC 7.3.1


sleiner:
The Travis build error seems unrelated:
```
CMake Error at /home/travis/bin/cmake/cmake-3.13.1-Linux-x86_64/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
660  Could NOT find Python3: Found unsuitable version "3.4.3", but required is
661  at least "3.6" (found /usr/bin/python3)
```

sleiner:
Just to be sure we are on the same page:
- `asm` is the keyword currently used in libmetal
- `__asm__` is the keyword that this pull request currently proposes, in line with:
  - GCC: https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords
  - Clang: https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes
  - The C section of cppreference.com: https://en.cppreference.com/w/c/language/asm

So in order for these changes to work with a compiler, it would need to support the `__asm__` keyword not `asm`
> It seems that some other compilers such as arm cc does not support **asm** ( to be confirmed).

sleiner:
I have no license for the Arm Compiler but from what I read in their [docs](https://developer.arm.com/documentation/100748/0614/Using-Assembly-and-Intrinsics-in-C-or-C---Code/Writing-inline-assembly-code?lang=en), they seem to support the very same GNU assembly syntax that GCC and Clang do, only with the `__asm` keyword instead of `__asm__`.

So my proposal would be to keep using `__asm__` in the code as it is the keyword endorsed by the two major compilers and work around keywords in other compilers like this:

```c
#ifdef IS_ARM_C_COMPILER
#define __asm__ __asm
#endif
```

This poses the question of how to check for the Arm Compiler. I have not found any compiler-specific set macros indicating this in their documentation...

sleiner:
Or were you talking about IAR? 
https://github.com/OpenAMP/libmetal/blob/dafc19ec8b86a4a9ae6e65d0a68a5434fb6d843d/lib/compiler.h#L17-L19

arnopo:
> Or were you talking about IAR?
> https://github.com/OpenAMP/libmetal/blob/dafc19ec8b86a4a9ae6e65d0a68a5434fb6d843d/lib/compiler.h#L17-L19

i 'm taking about diversity of the compilers. Today in openAMP only gcc and IAR are supported. But i know that some other compilers such as keil/arm cc are used.

If we focus on supported compiler:

For IAR the syntax is "asm": https://www.iar.com/support/resources/articles/expert-tip-about-intrinsics
i confirmed with a fast test in IAR that the syntax __asm__  generates a compilation error.

A alternative would be to replace "asm" by a macro such as "LIBMETAL_ASM" that would be defined in lib/compiler/<compiler>/compiler.h.

sleiner:
Oh, I realized just now why I did not understand your first message: The double underscores around a word are interpreted by the GitHub Markdown parser as bold, so if your message says **asm**, you really mean `__asm__` - that clears some things up 😄

> A alternative would be to replace "asm" by a macro such as "LIBMETAL_ASM" that would be defined in lib/compiler//compiler.h.

I thought about that too - but discarded it as the whole inline assembly syntax is GCC extension (although widely adopted across other compilers and available in the strict ISO modes). As far as I understand the concept of the `METAL_*` macros is to encapsulate non-standard behavior so that the code using the macros can be pure standard C. In this case, that is not possible by just aliasing the keyword. So using a `METAL_*` macro would hide the fact that the entire construct is a compiler extension.
However, I'm really not religious about this, so IMO you as the maintainer whether or not to introduce a `METAL_*` macro for this. What do you think?

Either way, fixing this for IAR is trivial - I can do it later today 👍

sleiner:
As for the Arm Compiler, it neither supports the current `asm` nor the proposed `__asm__` but its own `__asm`, so the current master branch should not compile with it unmodified anyway. So I see support for that compiler as a separate topic (which is hopefully addressed by someone with real access to that compiler).

sleiner:
@arnopo in case you did not see this - I'm waiting for your answer 😃 
> you as the maintainer whether or not to introduce a METAL_* macro for this. What do you think?

arnopo:
> @arnopo in case you did not see this - I'm waiting for your answer 😃
> 
> > you as the maintainer whether or not to introduce a METAL_* macro for this. What do you think?
This is an alternative yes.
Looking more in details, perhaps something more smart could be to keep asm in code and add asm definition  in lib/compiler/gcc/compiler.h . This would avoid to impact other compilers.
Could be something like that:

/* for C11 compatibility */
#if !defined( asm)
#define asm __asm__
#endif 

Of course if you have a compilation define related to  C11 compatibility to propose to replace "#if !defined( asm)" by it, don't hesitate

sleiner:
Alright, thanks 👍🏻 I will push a patch for IAR later and **not** introduce `METAL_ASM`.

sleiner:
When redefining the keyword, we can't lean on the C standard because `__asm__` (as all keywords beginning with `__` are) is part of the compiler implementer's namespace (unlike `asm` which is not supposed to be defined by standard implementers).

Also, we can't use something like `#if defined(__asm__)` to check for compiler support of the keyword, because the keywords are not known by the preprocessor anyway.

So, to summarize: If a keyword substitution is necessary can only be decided by checking for the specific compiler in use. That's why I put the substitution into IAR's `compiler.h` (without any further conditions).

sleiner:
Just saw this, the formatting in your answer was a bit off...

>Looking more in details, perhaps something more smart could be to keep asm in code and add asm definition in lib/compiler/gcc/compiler.h . This would avoid to impact other compilers.
What I don't like about that is that it lets non-standard code be accepted by the compiler and this also affects clients. They could be using the non-standard `asm` and compile even in strict modes, even though they are actually not standard-compliant.

To maximize compiler compatibility, one could also do something like:
```c
#ifndef STRICT_ISO_C_MODE // this macro does not really exist
#define __asm__ asm
#endif
```
But for that, you would have to know the compiler-specific defines for ISO C mode. So this is also not optimal.

All in all, as libmetal uses the non-standard inline assembly feature, there are dependencies on the features of specific compilers. So if anyone uses a non-supported compiler, they would have to write (and hopefully contribute) a custom compiler header. Looking at the compiler headers that exist now, this would not be a lot of effort.

arnopo:
I think we first need to determine what syntax is acceptable in strict C11 mode. It is not obvious to me that the use of `__asm__` is the only one .
The IRA also advertises compliance with C11 but does not support `__asm__`

It seems to me that it is only a gcc design choice. I have looked in the ISO/IEC 9899:201x document but nothing is related to the use of `__asm__`, `asm` is mentionned. But I am not sure I am referring to the right documents.

So if you have documents that detail C11 syntax (not related to gcc or clang), could you share them?

Please also have a look to __STDC_VERSION__ (https://sourceforge.net/p/predef/wiki/Standards/). Perhaps this could be used for gcc to define "asm".

sleiner:
That's the thing: Inline assembly is not part of ISO C at all but a compiler extension in any case. It was pioneered by GCC (with the `asm` keyword) and the other compilers copied it so you can use the same syntax across many compilers, like GCC, clang, IAR and the ARM C compiler.

In ISO C, compiler-specific keywords are prefixed by `__`, so that's why [GCC opted for `__asm__` in the strict ISO C modes](https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords), and [clang did too](https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes). [ARM CC](https://developer.arm.com/documentation/100748/0614/Using-Assembly-and-Intrinsics-in-C-or-C---Code/Writing-inline-assembly-code?lang=en) and [IAR CC (see page 166)](http://ftp.iar.se/WWWfiles/arm/webic/doc/EWARM_DevelopmentGuide.ENU.pdf) instead opted for `__asm`.

So you still need a compiler supporting the inline assembly extensions, but this pull request enables clients to use the strict ISO C modes so that their own code is properly checked for ISO C compliance.

sleiner:
I'm not really sure about `__STDC_VERSION__` - just like inline assembly this does not seem to be part of the ISO standard.

arnopo:
> That's the thing: Inline assembly is not part of ISO C at all but a compiler extension in any case. It was pioneered by GCC (with the `asm` keyword) and the other compilers copied it so you can use the same syntax across many compilers, like GCC, clang, IAR and the ARM C compiler.
> 
> In ISO C, compiler-specific keywords are prefixed by `__`, so that's why [GCC opted for `__asm__` in the strict ISO C modes](https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords), and [clang did too](https://clang.llvm.org/docs/UsersManual.html#differences-between-various-standard-modes). [ARM CC](https://developer.arm.com/documentation/100748/0614/Using-Assembly-and-Intrinsics-in-C-or-C---Code/Writing-inline-assembly-code?lang=en) and [IAR CC (see page 166)](http://ftp.iar.se/WWWfiles/arm/webic/doc/EWARM_DevelopmentGuide.ENU.pdf) instead opted for `__asm`.

So seems that replacing ` asm` by `__asm__` is not generic , right? 
In this case  i can see 2 alternatives:
- keep `asm' in code and find a way to define it for gcc with -std=c11 option
- go back to the `LIBMETAL_ASM` macro  

i would prefer the first solution keeping  `asm`. Could you try to use __STDC_VERSION__  for the gcc?  If it works i think is reasonnable to do it in this way. This would minimize impact for other compilers.
Then in future some fixes  will be added for IAR to support C11 if needed.

> 
> So you still need a compiler supporting the inline assembly extensions, but this pull request enables clients to use the strict ISO C modes so that their own code is properly checked for ISO C compliance.

sleiner:
I just tried with a small [test program](https://gist.github.com/sleiner/3cf61abe32df02a36eb1dc9c896ddd3c):
```bash
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c std-test.c         
std-test.c:7:9: note: #pragma message: 201710L
 #pragma message STR(__STDC_VERSION__)
         ^~~~~~~
std-test.c:15:9: note: #pragma message: __STRICT_ANSI__ not defined
 #pragma message "__STRICT_ANSI__ not defined"
         ^~~~~~~
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -std=c17 std-test.c  
std-test.c:7:9: note: #pragma message: 201710L
 #pragma message STR(__STDC_VERSION__)
         ^~~~~~~
std-test.c:13:9: note: #pragma message: 1
 #pragma message STR(__STRICT_ANSI__)
         ^~~~~~~
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -std=c11 std-test.c  
std-test.c:7:9: note: #pragma message: 201112L
 #pragma message STR(__STDC_VERSION__)
         ^~~~~~~
std-test.c:13:9: note: #pragma message: 1
 #pragma message STR(__STRICT_ANSI__)
         ^~~~~~~
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -std=gnu11 std-test.c
std-test.c:7:9: note: #pragma message: 201112L
 #pragma message STR(__STDC_VERSION__)
         ^~~~~~~
std-test.c:15:9: note: #pragma message: __STRICT_ANSI__ not defined
 #pragma message "__STRICT_ANSI__ not defined"
         ^~~~~~~
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -std=gnu89 std-test.c
std-test.c:9:9: note: #pragma message: __STDC_VERSION__ not defined
 #pragma message "__STDC_VERSION__ not defined"
         ^~~~~~~
std-test.c:15:9: note: #pragma message: __STRICT_ANSI__ not defined
 #pragma message "__STRICT_ANSI__ not defined"
         ^~~~~~~
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -ansi std-test.c   
std-test.c:9:9: note: #pragma message: __STDC_VERSION__ not defined
 #pragma message "__STDC_VERSION__ not defined"
         ^~~~~~~
std-test.c:13:9: note: #pragma message: 1
 #pragma message STR(__STRICT_ANSI__)
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -std=c89 std-test.c
std-test.c:9:9: note: #pragma message: __STDC_VERSION__ not defined
 #pragma message "__STDC_VERSION__ not defined"
         ^~~~~~~
std-test.c:13:9: note: #pragma message: 1
 #pragma message STR(__STRICT_ANSI__)
         ^~~~~~~
foo@bar ~/tmp % aarch64-none-elf-gcc-8.2.0 -c -std=c99 std-test.c
std-test.c:7:9: note: #pragma message: 199901L
 #pragma message STR(__STDC_VERSION__)
         ^~~~~~~
std-test.c:13:9: note: #pragma message: 1
 #pragma message STR(__STRICT_ANSI__)
         ^~~~~~~
```

So, to sum up: 
- `___STDC_VERSION__` is not defined in the modes up to C89 (GNU and ISO) and is defined from C99 and upward (GNU and ISO) which includes the current default.
- `__STRICT_ANSI__` is defined in all ISO modes and not defined in all the GNU modes (which include the default). So this is probably the check you were looking for.
- `asm` is available exactly if `__STRICT_ANSI__` is not defined
- `__asm__` is available in all tested versions

sleiner:
> In this case i can see 2 alternatives:
> 
> * keep `asm' in code and find a way to define it for gcc with -std=c11 option
> * go back to the `LIBMETAL_ASM` macro

I would prefer the `LIBMETAL_ASM` approach because it minimizes the impact on client code. If we define `asm` in the non-ISO mode, we're deliberately circumvent a compiler check, so if the following program compiles in the strict modes is dependent on whether a libmetal header is included:
```c
int main() { asm(""); }
```
I would argue that if these programs are compiled in strict modes, this should not be accepted in any case.

Using a special `LIBMETAL_ASM` macro on the other hand does not have that effect (client code using `LIBMETAL_ASM` would always compile and using `LIBMETAL_ASM` without the libmetal headers would fail, which is to be expected).
So I propose something like:
```c
#if defined(__STRICT_ANSI__)
#define LIBMETAL_ASM __asm__
#else
#define LIBMETAL_ASM asm
#endif
```
This would not break any builds that are working until now and it would not influence the ISO C checks for clients.

arnopo:
> > In this case i can see 2 alternatives:
> > 
> > * keep `asm' in code and find a way to define it for gcc with -std=c11 option
> > * go back to the `LIBMETAL_ASM` macro
> 
> I would prefer the `LIBMETAL_ASM` approach because it minimizes the impact on client code. If we define `asm` in the non-ISO mode, we're deliberately circumvent a compiler check, so if the following program compiles in the strict modes is dependent on whether a libmetal header is included:
> 
> ```c
> int main() { asm(""); }
> ```
> 
> I would argue that if these programs are compiled in strict modes, this should not be accepted in any case.
> 
> Using a special `LIBMETAL_ASM` macro on the other hand does not have that effect (client code using `LIBMETAL_ASM` would always compile and using `LIBMETAL_ASM` without the libmetal headers would fail, which is to be expected).

I have not strong opinion for this,  both alternatives LGTM.
As compilation generates an error if neither gcc compiler nor IAR compiler is used,  use the macro make also sense in term of legacy support.

Anyway I let @edmooring  gives his point of view.

> So I propose something like:
> 
> ```c
> #if defined(__STRICT_ANSI__)
> #define LIBMETAL_ASM __asm__
> #else
> #define LIBMETAL_ASM asm
> #endif
> ```
LGTM.
This has to be done in lib/compiler/gcc/compiler.h  and please define `asm` for IAR for legay compatibility. If needed a fix will be done for IAR  to  use __asm.
> 
> This would not break any builds that are working until now and it would not influence the ISO C checks for clients.

sleiner:
> This has to be done in lib/compiler/gcc/compiler.h and please define asm for IAR for legay compatibility.
Done 👍

arnopo:
Look good to me but please rework you commits  to squash them in two patches  to ease the review but also that we can recompile the library at each point of the git branch (in your series IAR compilation probably fails until the last commit). 
A first patch to define the macro and a second one to replace the `asm` in code would be sufficent.

Issue #101, opened 2019-10-10
FreeRTOS mutex does not handle priority inversion


supergaute:
Commit 10a0d5be removes the priority inheritance mechanism FreeRTOS is using with its mutexes.
Suggest to revert this commit.

edmooring:
The reasoning behind the original change is that FreeRTOS mutexes can't be statically initialized, which conflicts with the bare metal and Linux implementations. This will take some additional code restructuring.

supergaute:
Do you have any plans to do this restructuring in the near future?

arnopo:
@ed, @bentheredonethat,
I'm agree  with @supergaute. we should take benefit of FreeRTOS mutex. 

I'm trying to understand why the mutex can not be statically initialized using FreeRTOS. I'm not familiar with freertos, so don't hesitate if i missed something.
In freertos there is the following function: xSemaphoreCreateBinaryStatic [1]

This function seems to allow the possibility to have a static semaphore allocation
Therefore we could not define the METAL_MUTEX_INIT. As this macro is not used by the open-amp library, this should not affect it.

My proposal would be to  revert the 10a0d5b, and having  a static 

```
typedef struct {	typedef struct {
        StaticSemaphore_t semaphore;
	SemaphoreHandle_t m;	
} metal_mutex_t;

#define METAL_MUTEX_DEFINE(m) metal_mutex_t m;

static inline void __metal_mutex_init(metal_mutex_t *mutex)	static inline void __metal_mutex_init(metal_mutex_t *mutex)
{	{
	metal_assert(mutex);		atomic_store(&mutex->v, 0);
	mutex->m = xSemaphoreCreateBinaryStatic(mutex->semaphore);	
	metal_assert(mutex->m != NULL);	
}	}

```
I'm not able to test it,  so could someone validate if the implementation could work using FreeRTOS? 
 
[1] https://www.freertos.org/xSemaphoreCreateBinaryStatic.html

edmooring:
I'm inclined to revert the commit as well, and, since I can't find any evidence anybody is using METAL_MUTEX_INIT, deprecate it for this release, possibly with a compile time warning.

bentheredonethat:
@edmooring  @arnopo any update on this? if not,  I can send PR to revert the commit

arnopo:
I think you can send a PR to revert it, the current implementation seems not suitable for the freeRTOS. 
[...]

PR #126, opened 2020-07-31
Fix mutex implementation for FreeRTOS and Generic

sleiner:
The` atomic_flag_*` functions are only defined for operands of the type `atomic_flag` [1] so the previous implementation relied on undefined behavior.
Because of the need for reading the status without side effects (in `__metal_mutex_is_acquired`), it is not possible to use the `atomic_flags` type for the mutex implementation. Because of this, this pull request replaces all  `atomic_flag_*` functions with the appropriate functions for atomic value types.

[1] e.g. see: https://en.cppreference.com/w/c/atomic/atomic_flag_test_and_set

### DIFF
--- a/lib/system/freertos/mutex.h
+++ b/lib/system/freertos/mutex.h
@@ -17,7 +17,6 @@
 #define __METAL_FREERTOS_MUTEX__H__
 
 #include <metal/atomic.h>
-#include <stdlib.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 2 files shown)


arnopo:
LGTM but i need to perform extra tests using IAR, to ensure that this patch is compatible with the compiler.

arnopo (approved):
If this PR does not solve the https://github.com/OpenAMP/libmetal/issues/101, it seems to solve the problem related to the type of atomic flag faced by the IRA

PR #127, opened 2020-07-31
Support C++ clients using <atomic>

sleiner:
This has been tried (81f85dc244d05448e26aa674301065a1220b4539) and reverted (23c3c332acd7e649033ba63d8905df7d59c1540c) before. This pull request is a bit more elaborate than the first try:

- When the headers are compiled as part of a C++ file, the functions and types from `<atomic>` are exported into the global namespace so that the inline functions still work (ba233adeeb0ed881ddf6c91cba7d35f27d1e5b46)
- A few more explicit typecasts are introduced where needed by the C++ implementation of `atomic_store_explicit` (195d46b46ff0ec825a5b4ce25e45fe1fea085665)
- Because the `atomic_flag_*` C++ functions also perform stricter type checking than the C ones, this depends on #126

### DIFF
--- a/lib/atomic.h
+++ b/lib/atomic.h
@@ -14,8 +14,87 @@
 
 #include <metal/config.h>
 
-#if defined(HAVE_STDATOMIC_H) && !defined(__STDC_NO_ATOMICS__) && \
-	!defined(__cplusplus)
+#if defined(__cplusplus)
+# include <cstdint>
+
+/*
+ * <atomic> has the same functionality as <stdatomic.h> but all members are only
+ * accessible in the std namespace. As the rest of libmetal is pure C, it does
+ * not know about namespaces, even when compiled as part of a C++ file. So we
+ * just export the members of <atomic> into the global namespace.
+ */
+# include <atomic>
+using std::atomic_flag;
+using std::memory_order;
+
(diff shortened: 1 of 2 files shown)


sleiner:
I removed the duplicate commit.
However, the current HEAD will not compile with C++ clients because the fixes in #126 are needed for that. So it should probably be rebased again once #126 is merged.

arnopo:
> I removed the duplicate commit.
> However, the current HEAD will not compile with C++ clients because the fixes in #126 are needed for that. So it should probably be rebased again once #126 is merged.

Yes seems a limitation of Github. It would be nice to be able to describe dependencies in Pull requests...
We will request you a merge if needed.

@edmooring:
I don't have a strong opinion on this point. But, based on Zephyr's experience, I think that allowing commits duplication to carry out CI tests might be reasonable. If the sha1 commits are identical, the duplication should be resolved in the git merge, otherwise we will ask for a new base. 
Please give me your preference

arnopo (approved):

edmooring (approved):

edmooring:
@arnopo I'll update my work flow to accommodate these kind of changes more readily. Looks good to me.

Issue #117, opened 2020-05-04
Time unit mismatch in FreeRTOS __metal_sleep_usec


nsullivan33:
The math in sleep.h for FreeRTOS seems suspect, I think it might create delays 1,000 times larger than the caller expects:

static inline int __metal_sleep_usec(unsigned int usec)
{
	const TickType_t xDelay = usec / portTICK_PERIOD_MS;
	vTaskDelay(xDelay);
	return 0;
}

Shouldn't it be "(usec / 1000) / portTICK_PERIOD_MS" to convert microseconds to ticks?

arnopo:
Sorry for the late answer...
you are probably right regarding the APIs. Are you able to verify it in your environment?
Else you can propose the patch, we will find a way to verify it...

nsullivan33:
I can verify it and create a patch.  Do you need me to do both?  I think just dividing the "usec" variable by 1000, to convert to milliseconds, should be enough though the resolution would be poor.

arnopo:
If you can do both it is perfect. 
Concerning the calculation.  the precision will be one tick so dividing by 1000 is ok. 
And i think that we also add an extra  tick to the final result to ensure that we at least wait for the expected time..
"xDelay = ((usec / 1000) / portTICK_PERIOD_MS) + 1"
thanks!

arnopo:
@nsullivan33 
Do you plan to send a patch has you proposed?

nsullivan33:
Sorry, I've been incredibly busy lately.  I likely won't have time for a while...

Thanks,
Nathan Sullivan
________________________________

edmooring:
Taking a look at it:
The following small patch:

diff --git a/test/system/freertos/sleep.c b/test/system/freertos/sleep.c
index 9bc72d1..6964cd3 100644
--- a/test/system/freertos/sleep.c
+++ b/test/system/freertos/sleep.c
@@ -18,13 +18,14 @@
 static int sleep(void)
 {
        int rc;
-       unsigned int usec = 3;
+       unsigned int usec = 300000;
        unsigned long long tstart, tend, tdelayed;
 
        tstart = metal_get_timestamp();
        metal_sleep_usec((usec/portTICK_PERIOD_MS));
        tend = metal_get_timestamp();
        tdelayed = tend - tstart;
[...]

PR #129, opened 2020-10-03
FreeRTOS: fix sleep and timestamp handling

edmooring:
libmetal's libmetal_sleep_usec() takes a sleep time in micro-seconds.
The FreeRTOS function underlying it takes a sleep time in task ticks.
This was not properly corrected for, resulting in sleep times that
were off by an order of magnitude. The test code attempted to
correct for this as well, in the same wrong way.

Lastly, the metal_get_timestamp() function returned a value in
task ticks alone. This is a configuration dependent value.

This patch changes all three places to use microseconds for the
input and output values.

Addresses https://github.com/OpenAMP/libmetal/issues/117#

### DIFF
--- a/test/system/freertos/sleep.c
+++ b/test/system/freertos/sleep.c
@@ -18,14 +18,17 @@
 static int sleep(void)
 {
 	int rc;
-	unsigned int usec = 3;
-	unsigned long long tstart, tend, tdelayed;
+	unsigned int usec = 150000;
+	unsigned long long tstart, tend, tdelayed, tolerance;
+
 
 	tstart = metal_get_timestamp();
-	metal_sleep_usec((usec/portTICK_PERIOD_MS));
+	metal_sleep_usec(usec);
 	tend = metal_get_timestamp();
 	tdelayed = tend - tstart;
-	if (tdelayed > (usec/portTICK_PERIOD_MS))
+	/* allow +/- 1 OS tick */
+	tolerance = portTICK_PERIOD_MS * 1000;
+	if (tdelayed > (usec + tolerance) || tdelayed < (usec - tolerance))
 		rc = -1;
 	else
 		rc = 0;
--- a/README.md
+++ b/README.md
@@ -287,3 +287,5 @@ We use standard github mechanism for pull request. Please refer to github docume
 
 ## Communication and Collaboration
 [Subscribe](https://lists.openampproject.org/mailman/listinfo/openamp-rp) to the OpenAMP mailing list(openamp-rp@lists.openampproject.org).
+
+For more details on the framework please refer to the the [OpenAMP wiki](https://github.com/OpenAMP/open-amp/wiki).
--- a/lib/system/freertos/sleep.h
+++ b/lib/system/freertos/sleep.h
@@ -25,9 +25,9 @@ extern "C" {
 
 static inline int __metal_sleep_usec(unsigned int usec)
 {
-	const TickType_t xDelay = usec / portTICK_PERIOD_MS;
+	const TickType_t xDelay = ((usec/1000) / portTICK_PERIOD_MS);
 
-	vTaskDelay(xDelay);
+	vTaskDelay(xDelay ? xDelay : 1);
 	return 0;
 }
 
--- a/lib/system/freertos/time.c
+++ b/lib/system/freertos/time.c
@@ -15,6 +15,6 @@
 
 unsigned long long metal_get_timestamp(void)
 {
-	return (unsigned long long)xTaskGetTickCount();
+	return (unsigned long long)(1000 * portTICK_PERIOD_MS * xTaskGetTickCount());
 }
 


test/system/freertos/sleep.c
```diff
@@ -18,14 +18,14 @@
...
+	unsigned int usec = 10000;
 	unsigned long long tstart, tend, tdelayed;
 
 	tstart = metal_get_timestamp();
-	metal_sleep_usec((usec/portTICK_PERIOD_MS));
+	metal_sleep_usec(usec);
 	tend = metal_get_timestamp();
 	tdelayed = tend - tstart;
-	if (tdelayed > (usec/portTICK_PERIOD_MS))
+	if (tdelayed > (usec + 1000) || tdelayed < (usec - 1000))
```
arnopo, line 28: I wonder if this test is valid? I'm not sure I understand the value "1000"... Do you expect an accuracy of 1ms ? 
Look to me that tdelayed is a multiple of portTICK_PERIOD_MS so that you should use it in your test to avoid fails on some platforms.

Something like this:
if (tdelayed > (usec) || tdelayed < (usec + portTICK_PERIOD_MS)

test/system/freertos/sleep.c
```diff
@@ -18,14 +18,14 @@
 static int sleep(void)
 {
 	int rc;
-	unsigned int usec = 3;
+	unsigned int usec = 10000;
```
arnopo, line 21: This value corresponds to your  portTICK_PERIOD_MS, right?
I would suggest to use a non aligned value.

test/system/freertos/sleep.c
edmooring: The value 1000 was an arbitrary choice I made during my early investigation of the problem. You are correct. I will change the test to 
if (tdelayed > (usec + portTICK_PERIOD_MS * 1000) || tdelayed < (usec - portTICK_PERIOD_MS * 1000))
The reason behind the +/- 1 here is that depending on when in the tick period the sleep occurs, and how much usec is rounded by, the elapsed time could be 1 tick more or less than planned.

arnopo: > The value 1000 was an arbitrary choice I made during my early investigation of the problem. You are correct. I will change the test to
> if (tdelayed > (usec + portTICK_PERIOD_MS * 1000) || tdelayed < (usec - portTICK_PERIOD_MS * 1000))
> The reason behind the +/- 1 here is that depending on when in the tick period the sleep occurs, and how much usec is rounded by, the elapsed time could be 1 tick more or less than planned.

I think having an elapsed time lower than the one expected might be no safe. 
i would suggest to redesign the __metal_sleep_usec function, to ensure that the elapsed time is always higher than the expected one. Something like this:

```
static inline int __metal_sleep_usec(unsigned int usec)
{	{
        const TickType_t xDelay = ((usec/1000) / portTICK_PERIOD_MS);
        unsigned int remainder = usec % (1000 * portTICK_PERIOD_MS);

	vTaskDelay(remainder ? xDelay + 1 : xDelay);

	return 0;
}
```

edmooring: Since this is a wrapper around a function provided by FreeRTOS, I don't want to get too complicated. My concern is that we could accidentally always wait 1 tick too long because we try to compensate for something that is already being compensated for.

arnopo: vTaskDelay  seems not accurate. as explain in this post: https://www.freertos.org/FreeRTOS_Support_Forum_Archive/June_2011/freertos_vTaskDelay_problem_4570916.html

"If you ask for a 1 tick delay at a 1 ms tick, you will get somewhere between 0ms and 1ms, if you ask for a 2 tick delay, you will get a delay between 1ms and 2ms.  The low end of the range happens if you are at the end of a tick, and just after you call the delay function the tick occurs. Delays of course can also b extended if other tasks at your priority or higher become ready during the delay, as the delay just affect when this task is eligible to get CPU time, not when it will (unless it is the highest priority task) "

So in this case we have a chance to not respect the requested time. For instance requesting 19 ms for a 10 ms tick can lead to 0 ms in bad case.
That's why I would be more favorable to the idea of waiting 1 notch too much not to accumulate the inaccuracies of the division and vTaskDelay . 

That said your point is valid: This should only  be a wrapper from ms to tick.
Perhaps just implementing vTaskDelay( xDelay + 1 )  for all cases  would be a good compromise to compensate the truncated division?

edmooring: The FreeRTOS documentation confirms that this is the intended behavior. My changes will ensure that vTaskDelay() will never be called with a value of 0, which means something specific to FreeRTOS. I don't want to arbitrarily add 1 or do any rounding to keep things simple.

It seems that this is an area of libmetal needs a deeper look at the design and documentation. Can we use this FreeRTOS-specific change for the release and open an issue for a deeper look across the other implementations?

arnopo: > The FreeRTOS documentation confirms that this is the intended behavior. My changes will ensure that vTaskDelay() will never be called with a value of 0, which means something specific to FreeRTOS. I don't want to arbitrarily add 1 or do any rounding to keep things simple.

The +1 is not arbitrary here it is used to compensate the error introduced by the conversion from ms to tick, so not specific to FreeRTOS but to conversion in tick. This is something quite common for such conversion to respect at minimum the requested time.

In fact you already do it for the 0 value to avoid that vTaskDelay() returns immediately. 
This probably respond to the use case in rpmsg_virtio (https://github.com/OpenAMP/open-amp/blob/master/lib/rpmsg/rpmsg_virtio.c#L322)
RPMSG_TICKS_PER_INTERVAL is equal to 1ms ( with a tick at 5 or  10 ms).
So what i proposed here was to not add an exception for 0 only but to compensate for all the requested values.

That's said, examples in FreeRTOS do not add 1 to the conversion result.
So if you are still not convince ,  OK to merge this.

> It seems that this is an area of libmetal needs a deeper look at the design and documentation. Can we use this FreeRTOS-specific change for the release and open an issue for a deeper look across the other implementations?

test/system/freertos/sleep.c
edmooring: Changed in the updated commit.

test/system/freertos/sleep.c
edmooring: I would prefer not to add 1 in all cases, since that could result in the function taking 1 extra clock tick.

arnopo (changes requested):
Ok, let's do it in this way.  The most important here is to fix the missing conversion from microsecond to millisecond. 

Could you just fix issues reported bu checkpatch:

#48: FILE: lib/system/freertos/time.c:18:
+	return (unsigned long long)(1000 * portTICK_PERIOD_MS * xTaskGetTickCount());

129.patch:76: ERROR:MISSING_SIGN_OFF: Missing Signed-off-by: line(s)

arnopo (approved):
Seems that you branch is based on v2020.04 release and not on master, 
I guess that the rebase should work, and clean this.

PR #131, opened 2020-10-09
migrate CI build test from travis to github action 

arnopo:
This PR:

- creates a Github action to execute build for a specified target. This action replaces Travis test and so suppresses the dependency with Travis environment.
- fixes the broken Zephyr test
- splits the workflow in two parts to separate the build end non reg test with the code compliance tests.


arnopo:
Hi @galak
Please could you have a look on zephyr part of this PR and tell me if it is coherent and maintainable from your point of view?
Thanks in advance

edmooring (changes requested):
Otherwise, LGTM.

edmooring (approved):

PR #132, opened 2020-10-13
CI: rework compliance test

arnopo:
The webispy/checkpatch-action does not seem to work properly.
Even if it detects coding style issue, the return status is successful.
This patch replaces the action by the compliance python script obtained
from the Zephyr project.
This script also runs gitlint to check commit messages.

Notice that this PR depends on https://github.com/OpenAMP/libmetal/pull/131 
Only last commit named CI: "rework compliance test" as to be reviewed
The PR will we rebased when https://github.com/OpenAMP/libmetal/pull/131  will be integrated

Signed-off-by: Arnaud Pouliquen arnaud.pouliquen@st.com

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -18,12 +18,61 @@ on:
       - scripts/**
 
 jobs:
-  checkpatch_review:
-    name: checkpatch review
+  compliance_review:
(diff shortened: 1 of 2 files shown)


edmooring (approved):
Looks good to me.

PR #133, opened 2020-10-15
Set CMake minimum version to 3.0.2

edmooring:
Signed-off-by: Ed Mooring <ed.mooring@xilinx.com>


arnopo (changes requested):
Commit description is missing

arnopo (approved):

PR #134, opened 2020-10-16
Xilinx: Make xilmem and xilstandalone library linkage conditional.

edmooring:
In v2019.1, the xilmem and xilstandalone libraries were removed from
the standalone Xilinx BSP for the R5. This change adds a CMake
variable, XILINX_PRE_V2019, to conditionally add these libraries
to the link command for use with older BSPs.

Signed-off-by: Ed Mooring <ed.mooring@linaro.org>


examples/system/generic/zynqmp_r5/CMakeLists.txt
```diff
@@ -5,8 +5,10 @@ collect(PROJECT_LIB_DEPS c)
 collect(PROJECT_LIB_DEPS m)
 
 set (_lib "xil")
+if (XILINX_PRE_V2019)
```
arnopo, line 8: This look like a deprecation, so we probably should be treat it in this way.
 what about adding a warning message here:
 message( DEPRECATION " Compatibility for Xilinx 2018.3 or earlier environments, could be no more supported in future release" )

edmooring: This isn't really a deprecation. In release 2019.1, Xilinx combined the xilmem and xilstandalone libraries into the xil library, rather than having them as separate libraries. Ideally, OpenAMP should be capable of supporting older environments for a very long time.

arnopo: You are right it is not a API deprecation.
But If we do not apply similar deprecated process for the Xilinx platform we have also to apply same rule for other platforms and OS . My concern is that this could become not sustainable .
Support this during 2 year would seem to me more reasonable. After 2 years Xilinx platforms using release older than  2019.1 will be able to continue to use OpenAMP exemple but not to update to the last release.
Do you think that 2018 Xilinx releases will need to be compatible with 2022.10 release?

edmooring: One of the folks on the mailing list was using v2017.1 last year. Academic, aviation, and automotive users frequently keep their hardware and software environments for a long time.  One of my objectives with this change is to reduce the barriers to entry for new users, who may very well be using an old platform. I think a minimum of 5 years of compatibility with older platforms is probably about right. For example, someone using Ubuntu 16 LTS should still be able to compile OpenAMP in 2021.

Because this is a sample application, and is vendor-specific, the vendor should bear the burden of maintenance. I would not expect changes like these to happen too often.

arnopo: > Because this is a sample application, and is vendor-specific, the vendor should bear the burden of maintenance. I would not expect changes like these to happen too often.
Sold like this I can only approve it :)

arnopo (approved):

PR #135, opened 2020-10-21
CI: fix compliance check false report when merge a PR.

arnopo:
When a PR is merged, a "Run failed: libmetal compliance checks - master"
mail is generated.
As it is not necessary to rerun this compliance check (already done
on sent PR), just suppress the action for push event.

Signed-off-by: Arnaud Pouliquen <arnaud.pouliquen@st.com>


edmooring (approved):
Looks good to me.

PR #137, opened 2020-11-19
Linux: fix misunderstanding of the offset devicetree node in the map

lslrt:
I think there is a confusion about the of node "offset" of the uio map regions.
According to the user io documentation (https://www.kernel.org/doc/html/latest/driver-api/uio-howto.html), the offset node is 

> The offset, in bytes, that has to be added to the pointer returned by mmap()

this is NOT the offset to be passed to mmap.

The offset to give to mmap to get the Nth region is N times the pagesize as explained in the same documentation:

> From userspace, the different mappings are distinguished by adjusting the offset parameter of the mmap() call. To map the memory of mapping N, you have to use N times the page size as your offset:

```
offset = N * getpagesize();
```

The first patch https://github.com/OpenAMP/libmetal/pull/137/commits/2219961d1dc73d2f3e484f2701cf8fa18ebde1c7 of this series correct this issue.  

For the second https://github.com/OpenAMP/libmetal/pull/137/commits/357afeb7ae75d88dc3ae1aa8f6c06cd1803f48cc , that is tagged as [DO NOT MERGE], I'm not sure that adding a new field in the common structure `metal_io_region` that will be used only for linux is legit. Maybe we rather need to handle this in the `metal_io_ops`. I'm not familiar enough with libmetal, maybe someone with more insight can give a thought...
I do not have a lot of time now, and since I do not have the "Non page aligned" use case I cannot test easily, but I may find some time to produce a test bench for this scenario in the future.
What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,11 @@ implementation for metal_sys_init and metal_sys_finish.
 For Linux userspace, metal_sys_init sets up a table for available shared pages,
 checks whether UIO/VFIO drivers are avail, and starts interrupt handling
 thread.
+Please note that on Linux, to access device's memory that is not page
+aligned, an offset has to be added to the pointer returned by
+mmap(). This `offset`, although it can be read from the device tree
+property exposed by the uio driver, is not handled yet by the
+library.
 
 For bare-metal, metal_sys_init and metal_sys_finish just returns.
 
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -234,7 +234,7 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
 		result = (result ? result :
 			 metal_uio_read_map_attr(ldev, i, "size", &size));
 		result = (result ? result :
-			 metal_map(ldev->fd, offset, size, 0, 0, &virt));
+			 metal_map(ldev->fd, i * getpagesize(), size, 0, 0, &virt));
 		if (!result) {
 			io = &ldev->device.regions[ldev->device.num_regions];
 			metal_io_init(io, virt, phys, size, -1, 0, NULL);


lslrt:
@arnopo , @edmooring : did you had time to have a quick look? Maybe we need to open an issue for the patch https://github.com/OpenAMP/libmetal/commit/2219961d1dc73d2f3e484f2701cf8fa18ebde1c7 ?

lib/io.c
```diff
@@ -10,7 +10,7 @@
 #include <metal/sys.h>
 
 void metal_io_init(struct metal_io_region *io, void *virt,
-	      const metal_phys_addr_t *physmap, size_t size,
+	      const metal_phys_addr_t *physmap, size_t size, size_t offset,
```
edmooring, line 13: This changes a publicly facing API which we can't really do.

It's also incomplete, in that it doesn't actually use the offset anywhere. 

This is definitely a valid issue, but I believe it will require more thought and architecture to make libmetal use the offset properly. I think we can probably do something by adding an element to the libmetal Linux device structure and changing the way the memory region is initialized.

edmooring (changes requested):

lib/io.c
lslrt: I agree that this change cannot be taken as is, as explained it was only a way to initiate a talk.

> It's also incomplete, in that it doesn't actually use the offset anywhere.

Hum... maybe I've missed something, but please have a look at the metal_io_virt() function that is modified by the same patch where the new 'offset' field of the metal_io_region is used.

> This is definitely a valid issue, but I believe it will require more thought and architecture to make libmetal use the offset properly.

Should I make a new pull request with only the first patch that fix the issue + adding a note in the documentation saying that on Linux, devices that use memory that is not page aligned may not work?

lib/io.h
```diff
@@ -131,8 +133,8 @@ static inline size_t metal_io_region_size(struct metal_io_region *io)
 static inline void *
 metal_io_virt(struct metal_io_region *io, unsigned long offset)
 {
-	return (io->virt != METAL_BAD_VA && offset < io->size
-		? (uint8_t *)io->virt + offset
+	return (io->virt != METAL_BAD_VA && (offset + io->offset) < io->size
+		? (uint8_t *)io->virt + offset + io->offset
```
lslrt, line 137: @edmooring Here the new offset is used

edmooring: Cool. Sorry I missed that.

lib/io.c
edmooring: That would be much appreciated. 

We should also open an issue to track the problem of actually using the offset.

lslrt: @edmooring Done. Let me know what you think about the note that I've added into the README.md file

lib/system/linux/device.c
```diff
@@ -234,7 +234,7 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
 		result = (result ? result :
 			 metal_uio_read_map_attr(ldev, i, "size", &size));
 		result = (result ? result :
-			 metal_map(ldev->fd, offset, size, 0, 0, &virt));
+			 metal_map(ldev->fd, i * getpagesize(), size, 0, 0, &virt));
```
lslrt, line 237: `offset` (line 152) variable is not used any more, should I remove it? It may failed to build if -Werror and -Wunused compilation flag are turned on together.

README.md
```diff
@@ -154,6 +154,11 @@ implementation for metal_sys_init and metal_sys_finish.
 For Linux userspace, metal_sys_init sets up a table for available shared pages,
 checks whether UIO/VFIO drivers are avail, and starts interrupt handling
 thread.
+Please note that on Linux, to access device's memory that is not page
+aligned, an offset has to be added to the pointer returned by
+mmap(). This `offset`, although it can be read from the device tree
+property exposed by the uio driver, is not handled yet by the
+librarie.
```
lslrt, line 161: typo: librarie->library

lib/io.h
lslrt: Removed

lslrt (commented):
Hello,

@edmooring what do you think about this note?
What about the merge request?
Do I need to open a github issue to track the bug?

Thanks,

lib/io.c
lslrt: Removed

edmooring (approved):
Looks good to me.

lslrt:
@arnopo  Does it looks good for you too?

arnopo (approved):

PR #138, opened 2020-11-20
Improve compilation with IAR environment

arnopo:
Improve libmetal to better support  EWARM compiler used by  the IAR system workbench for Arm platform.


edmooring (approved):
Looks good to me.

PR #147, opened 2021-04-13
Version management update

arnopo:
As the libmetal is now quite stable,  a proposition is to create the V1 release. and then to follow [semver](https://semver.org/) rules.

In this scope the pull request:
- update the readme to provide a reference to the semver
- create a VERSION file to make it easier to find and update the version.
- rework version management in consequence, but also to use cmake version variables.

As a second step the open-amp library will be updated using same approach.


VERSION
```diff
@@ -0,0 +1,3 @@
+VERSION_MAJOR = 0
```
galak, line 1: Should MAJOR not be 1?

arnopo: Hi @galak
I would prefer to make it in two steps to set the version update in the last patch of the release.

galak (approved):

edmooring (approved):
Looks good to me.

PR #148, opened 2021-04-23
re-enable the use of Freertos mutexes

arnopo:
Revert the commit 10a0d5b according to the https://github.com/OpenAMP/libmetal/issues/101 discussions.

This can have impact on the METAL_MUTEX_INIT use and memory allocation for application.

But on the other hand, using the atomic mechanism for the semaphore bypasses the FreeRTOS priority-based preemptive scheduling policy, resulting in a possible deadlock in the OpenAMP library.


lib/system/freertos/mutex.h
```diff
@@ -16,24 +16,24 @@
...
 
-#define METAL_MUTEX_UNLOCKED 0
-#define METAL_MUTEX_LOCKED   1
-
 /*
  * METAL_MUTEX_INIT - used for initializing an mutex element in a static struct
  * or global
  */
-#define METAL_MUTEX_INIT(m)	{ ATOMIC_VAR_INIT(METAL_MUTEX_UNLOCKED) }
+#define METAL_MUTEX_INIT(m) { NULL }
```
edmooring, line 36: This will likely create code that compiles, but does not work in some unexpected ways, since there will be a mutex that is uninitialized.
We should at least warn the user about this. For gcc, we could do something like:
#define METAL_MUTEX_INIT(m) _Pragma("GCC warning \"Static initialization of mutexes is unsupported.\");

edmooring: On second thought. It might be better to treat it as an error. s/warning/error/.

edmooring (changes requested):

lib/system/freertos/mutex.h
arnopo: That's a good point. 
Something that it makes the picture more complex is that is macro is used to initialize a structure on declaration, as done in the  METAL_MUTEX_DEFINE macro.

I had a look for each supported compiler  GCC, EWARN and ARMCC, 
- GCC support the warning message that can be changed in error
- EWARM (IAR) allows only to print a message or an error
- not possible to use _Pragma for print a warning.

il will send an extra patch on top

arnopo:
Here is the checkpatch fails :

```
ERROR   : Test checkpatch failed: -:33: ERROR:MULTISTATEMENT_MACRO_USE_DO_WHILE: Macros with multiple statements should be enclosed in a do - while loop
#33: FILE: lib/system/freertos/mutex.h:37:
+#define METAL_MUTEX_INIT(m) { NULL }; \
+_Pragma("GCC warning\"static initialisation of the mutex is deprecated\"")
```
the error is linked to the commit https://github.com/OpenAMP/libmetal/pull/148/commits/8fab7f9aa9860a662a1cbdda9597c07ad774c7e1.
As the expected user code is `metal_mutex_t m = METAL_MUTEX_INIT(m)`  or `METAL_MUTEX_DEFINE(m)` it is not possible here to use  a do - while loop.

Unless someone has an alternative, we should probably ignore the checkpatch report here, if we want to generate a compile-time warning...

lib/system/freertos/mutex.h
edmooring: This causes libmetal-test-static to fail intermittently and inconsistently on the Xilinx R5. I think we can keep this open and fix it, but this change shouldn't go into 2021.04 as is.

arnopo: > This causes libmetal-test-static to fail intermittently and inconsistently on the Xilinx R5. I think we can keep this open and fix it, but this change shouldn't go into 2021.04 as is.

Agree, not reasonnable to integrate it without understanding and fixing the issue.

PR #154, opened 2021-05-03
cmake: fix VERSION file path

arnopo:
Regression found when integrating release in Zephyr project. 

cmake fails when CMAKE_SOURCE_DIR variable does not point
to the libmetal directory.

Fix by defining LIBMETAL_ROOT_DIR in the main CmakeLists.txt.

This fix https://github.com/OpenAMP/libmetal/commit/ad87802d6e01e97946de20b6c2fa28aed184ed20


galak (approved):

edmooring (approved):
Works for me.

PR #160, opened 2021-05-21
linux: device: unmap regions for linux device

lkolbly:
This unmaps the device regions which were opened during metal_uio_dev_open. Failure to do this results in a file descriptor leak.


arnopo (approved):
LGTM, thanks!

edmooring (approved):
Good catch! Thanks. Ready to go.

PR #162, opened 2021-05-29
io: add null pointer check in io block ops

PeikanTsai:
metal_io_virt returns NULL on invalid input metal_io_region.
Check ptr before use it.

Signed-off-by: Mark-PK Tsai <mark-pk.tsai@mediatek.com>

### DIFF
--- a/lib/io.c
+++ b/lib/io.c
@@ -39,7 +39,7 @@ int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 	unsigned char *dest = dst;
 	int retlen;
 
-	if (offset >= io->size)
+	if (!ptr)
 		return -ERANGE;
 	if ((offset + len) > io->size)
 		len = io->size - offset;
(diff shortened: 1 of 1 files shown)


PeikanTsai:
@arnopo  @edmooring  could you please review it?

lib/io.c
```diff
@@ -39,7 +39,7 @@ int metal_io_block_read(struct metal_io_region *io, unsigned long offset,
 	unsigned char *dest = dst;
 	int retlen;
 
-	if (offset >= io->size)
+	if (!ptr || offset >= io->size)
```
arnopo, line 42: (offset >= io->size) is already tested in metal_io_virt 
So testing !ptr should be sufficient

arnopo (changes requested):
The point is valid, thanks 
LGTM just can be optimized as offset >= io->size is already tested in metal_io_virt

lib/io.c
PeikanTsai: Agree. I'll update this patch to remove it as you suggested.

arnopo (approved):
LGTM, thanks!

edmooring (approved):
Looks good to me.

PR #163, opened 2021-06-08
zephyr:  support of the Cortex-A and Cortex-R (AArch32)

julien-massot:
Cortex A-M-R  have the same WFI assembly instruction
and so can works with the existing cortexm support.

Rename cortexm directory to 'arm'.

Build tested with stm32mp157c_dk2 for cortex-m,
rcar_h3ulcb for cortex-r.

Signed-off-by: Julien Massot <julien.massot@iot.bzh>


cmake/syscheck.cmake
```diff
@@ -7,7 +7,12 @@ if (WITH_ZEPHYR)
   if (NOT WITH_ZEPHYR_LIB)
     include($ENV{ZEPHYR_BASE}/cmake/app/boilerplate.cmake NO_POLICY_SCOPE)
   endif()
-  if (CONFIG_CPU_CORTEX_M)
-    set (MACHINE "cortexm" CACHE STRING "")
-  endif (CONFIG_CPU_CORTEX_M)
+  if (CONFIG_CPU_CORTEX_A)
+    set (MACHINE "cortex_a_m_r" CACHE STRING "")
```
arnopo, line 11: what about just "arm" ?

julien-massot: yes that's better, looking for a better CONFIG_ to check, CONFIG_ARM maybe a good candidate.

edmooring (approved):
Looks sane to me.

arnopo (approved):
LGTM

@galak, @ioannisg : Since this PR is related to the zephyr environment, could  you confirm that it is compliante with the ARM configs management in zephyr?

Thanks!

galak (approved):

cmake/syscheck.cmake
```diff
@@ -7,7 +7,8 @@ if (WITH_ZEPHYR)
   if (NOT WITH_ZEPHYR_LIB)
     include($ENV{ZEPHYR_BASE}/cmake/app/boilerplate.cmake NO_POLICY_SCOPE)
   endif()
-  if (CONFIG_CPU_CORTEX_M)
-    set (MACHINE "cortexm" CACHE STRING "")
-  endif (CONFIG_CPU_CORTEX_M)
+  if (CONFIG_ARM)
```
ioannisg, line 10: That's correct, since ARM signifies, actually AARC32. So it does not enable AARC64. 
That's just an FYI, and perhaps we need to change the PR title/body.

ioannisg (approved):

cmake/syscheck.cmake
julien-massot: Thanks, it's done !

PR #176, opened 2021-10-15
test: zephyr: fix 'errno' undeclared

arnopo:
The return value should be negative.
Added inclusion of errno.h to avoid an unresolved symbol
error when linking the  test.


edmooring (approved):
Looks good to me.

PR #184, opened 2021-11-05
libmetal: Compiles for petalinux microblazeel.

thesamprice:
Not sure if this is needed / wanted.
I got libmetal to compile for microblaze 2020.2 by adding the atomic library to required libs.
I also had to copy the folder from microblaze to microblazeel.

I built by
source /opt/petalinux/2020.2/environment-setup-microblazeel-v11.0-bs-cmp-re-mh-div-xilinx-linux
mkdir -p libmetal/build
cd libmetal/build
cmake ..
make

### DIFF
--- a/lib/processor/microblazeel/CMakeLists.txt
+++ b/lib/processor/microblazeel/CMakeLists.txt
@@ -1,0 +1,4 @@
+collect (PROJECT_LIB_HEADERS atomic.h)
+collect (PROJECT_LIB_HEADERS cpu.h)
+collect (PROJECT_LIB_DEPS atomic)
+# vim: expandtab:ts=2:sw=2:smartindent
--- a/lib/processor/microblazeel/atomic.h
+++ b/lib/processor/microblazeel/atomic.h
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2017, Xilinx Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	microblaze/atomic.h
+ * @brief	Microblaze specific atomic primitives for libmetal
+ */
+
+#ifndef __METAL_MICROBLAZE_ATOMIC__H__
+#define __METAL_MICROBLAZE_ATOMIC__H__
+
+#endif /* __METAL_MICROBLAZE_ATOMIC__H__ */
--- a/lib/processor/microblazeel/cpu.h
+++ b/lib/processor/microblazeel/cpu.h
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2017, Xilinx Inc. and Contributors. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	cpu.h
+ * @brief	CPU specific primatives on microblaze platform.
+ */
+
+#ifndef __METAL_MICROBLAZE__H__
+#define __METAL_MICROBLAZE__H__
+
+#include <stdint.h>
+#include <metal/atomic.h>
+
+#define metal_cpu_yield()
+
+#endif /* __METAL_MICROBLAZE__H__ */


lib/processor/microblazeel/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+collect (PROJECT_LIB_HEADERS atomic.h)
+collect (PROJECT_LIB_HEADERS cpu.h)
+collect (PROJECT_LIB_DEPS atomic)
+# vim: expandtab:ts=2:sw=2:smartindent
```
arnopo, line 4: to remove line

lib/processor/microblazeel/atomic.h
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (c) 2017, Xilinx Inc. and Contributors. All rights reserved.
```
arnopo, line 2: copyright not up to date

lib/processor/microblazeel/cpu.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2017, Xilinx Inc. and Contributors. All rights reserved.
```
arnopo, line 2: same here

arnopo (changes requested):
Please fix commit message to respect rule ( commit subject ansd signature).
More important could explain why you can not reuse the microblaze?
I also made few comments on code

thesamprice:
@arnopo
My normal build process is to source the petalinux toolchain.
cat /opt/petalinux/2020.2/environment-setup-microblazeel-v11.0-bs-cmp-re-mh-div-xilinx-linux
They spec 2 environment variables when sourcing this.
The libmetal was looking for microblazeel, so i adjusted the folder to fix.

```
...
export ARCH=microblaze
export OECORE_TARGET_ARCH="microblazeel"
...
```

Looks like libmetal uses `PROJECT_PROCESSOR`, which maybe i could build a custom toolchain file to set.
I couldnt find where that was set though with a quick search.
Out of the box build using the petalinux setup environment didnt work for me.
Maybe its better to request petalinux toolchain to set the correct cmake variables for target processors so toolchains work better.

Let me know how you want to handle this.

arnopo:
> @arnopo My normal build process is to source the petalinux toolchain. cat /opt/petalinux/2020.2/environment-setup-microblazeel-v11.0-bs-cmp-re-mh-div-xilinx-linux They spec 2 environment variables when sourcing this. The libmetal was looking for microblazeel, so i adjusted the folder to fix.
> 
> ```
> ...
> export ARCH=microblaze
> export OECORE_TARGET_ARCH="microblazeel"
> ...
> ```
> 
> Looks like libmetal uses `PROJECT_PROCESSOR`, which maybe i could build a custom toolchain file to set. Gets set in libmetal/cmake/options.cmake from CMAKE_SYSTEM_PROCESSOR
> 
> Out of the box build using the petalinux setup environment didnt work for me. Maybe its better to request petalinux toolchain to set the correct cmake variables for target processors so toolchains work better.
> 
> Let me know how you want to handle this.

Did you try to use this cmake platform file ( or create a new one based on it ?
https://github.com/OpenAMP/libmetal/blob/master/cmake/platforms/microblaze-generic.cmake

`cmake .. -DCMAKE_TOOLCHAIN_FILE=../microblaze-generic.cmake`

thesamprice:
I did not use that one. I think that ones for the stand alone / baremetal build of microblaze? 

im building for linux running on micro-blaze via the petalinux method.
I tried building it via petalinux-config -c rootfs/yocto, and that seems broken for microblaze so tried doing it this method.

I have a modified toolchain file where i was setting up the cmake sys root, but libmetal wasn't finding pthreads library.

Not sure if im doing this wrong, or not.  It built so next is to get the elf loader working

I feel like running openamp / libmetal on microblaze linux is a new thing?
Was hoping to prototype on this, and then see how hard it would be to port to rtems

edmooring:
@thesamprice Building OpenAMP/libmetal on Microblaze Linux is not part of the usual dev workflow. That being said, did you try enabling packagegroup-petalinux-openamp in your rootfs?

thesamprice:
@edmooring
I tried modifying `project-spec/meta-user/conf/user-rootfsconfig`

and adding `CONFIG_libmetal` and then enabling in the petalinux-config -c rootfs

 However when building petalinux it failed with an undefined variable ${SOC_FAMILY_ARCH}.
 
 I then went via the method described above to build the static library.  Although i imagine parts of it need to be moved into a kernel module.

arnopo:
@thesamprice 
What is the status, can we close this PR?

thesamprice:
> 

Yea go ahead and close it.  I was waiting on a reply from @edmooring to determine if i should move forward or not.
I probably wont get back into openamp until im ready to port my elf loader to rtems.

r-kyle15:
@arnopo, @edmooring 
Came across this issue as well, I checked that meta-openamp was included in bblayers.conf and added CONFIG_libmetal to rootfs.  At least for the microblaze, the packagegroup-petalinux-openamp does not exist in the rootfs config menu.  The only thing that allowed a successful build was following @thesamprice suggested edits to the lib/processor folder.  Recommend pulling in updates so future users can avoid this issue.

arnopo:
> @arnopo, @edmooring Came across this issue as well, I checked that meta-openamp was included in bblayers.conf and added CONFIG_libmetal to rootfs. At least for the microblaze, the packagegroup-petalinux-openamp does not exist in the rootfs config menu. The only thing that allowed a successful build was following @thesamprice suggested edits to the lib/processor folder. Recommend pulling in updates so future users can avoid this issue.

I'm not familiar with the Xilinx environment, @edmooring has a better knowledge of the Xilinx platforms.

That's said from my windows I can see 3 layers involved:
- Yocto build environment (Xilinx one?)
- meta-openamp layer (Xilinx one?)
- libmetal
The question is in which layer(s) this point  should be addressed.

From libmetal perspective:
What does the "el" in "microblazeel" mean?
Could we consider here that the processor is the microblaze and the system is Linux?  

In this case creating  a `cmake/platforms/microblaze-linux.cmake` file to support Linux on Microblaze processor could make sense.

Then concerning the  meta-openamp layer and Yocto build: 
Seems that it is Xilinx specific, right?
I suppose that some adaptations are possible also possible in these layers ...

thesamprice:
el stands for endian little as much as i can figure out.

I remember the libmetal was picking up on the cmake variables that are set via the stock environment variables from sourcing the petalinux toolchain file.
`/opt/Xilinx/PetaLinux/2020.2/environment-setup-microblazeel-v11.0-bs-cmp-re-mh-div-xilinx-linux`
Sets up a bunch of stuff specifically naming the arch as "microblazeel.
```
export OECORE_TARGET_ARCH="microblazeel"
export OECORE_TARGET_OS="linux"
```
i imagine this could be overcome with a better toolchain file.

arnopo:
I am closing this PR as it does not seem to be strongly requested and has been stale for a long time. Feel free to reopen it, if still needed, addressing comments.

PR #185, opened 2021-11-07
metal_init: Add a flag to indicate that internal data is initialized.

edmooring:
Linux users have been seeing segfaults in libmetal when metal_finish()
is called twice after a successful initialization, or on exit afer
an unsuccessful initialization. This adds a field to metal_common_state
that is set on initialization. metal_finish() sets the entire structure,
including this field to zero. metal_finish() now also checks this flag
before attempting to shutdown libmetal.

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -24,11 +24,15 @@ int metal_init(const struct metal_init_params *params)
 	if (error)
 		return error;
 
+	_metal.common.initialized = 1;
 	return error;
 }
 
 void metal_finish(void)
 {
+	/* if metal_init() failed, or metal_finish() is called twice, just return */
+	if (!_metal.common.initialized)
+		return;
 	metal_sys_finish();
 	memset(&_metal, 0, sizeof(_metal));
 }
--- a/lib/sys.h
+++ b/lib/sys.h
@@ -67,6 +67,9 @@ struct metal_common_state {
 	/** Current log handler (null for none). */
 	metal_log_handler		log_handler;
 
+	/** Has the structure been initialized? */
+	int initialized;
+
 	/** List of registered buses. */
 	struct metal_list		bus_list;
 


lib/init.c
```diff
@@ -24,11 +24,15 @@ int metal_init(const struct metal_init_params *params)
...
 		return error;
 
+	_metal.common.initialized = 1;
 	return error;
 }
 
 void metal_finish(void)
 {
+	/* if metal_init() failed, or metal_finish() is called twice, just return */
+	if (_metal.common.initialized)
```
arnopo, line 34: I miss something or should be `if(! _metal.common.initialized)`?

arnopo (commented):
I supposse that the  segfault occurs in metal_sys_finish, right?
Could it be fix in metal_sys_finish in this case?

lib/init.c
edmooring: It definitely should be. It was in my source tree, but not in git. Thanks for the catch.

edmooring:
Let's see if I got it right this time.

arnopo:
Hi @edmooring 


> I supposse that the segfault occurs in metal_sys_finish, right?
> Could it be fix in metal_sys_finish in this case?

What aboout fixing it in https://github.com/OpenAMP/libmetal/blob/master/lib/system/linux/init.c#L168?
Seems to me that the segfault is linked to Linux implementation.

lib/init.c
```diff
@@ -24,11 +24,15 @@ int metal_init(const struct metal_init_params *params)
...
 		return error;
 
+	_metal.common.initialized = 1;
 	return error;
 }
 
 void metal_finish(void)
 {
+	/* if metal_init() failed, or metal_finish() is called twice, just return */
+	if (!_metal.common.initialized)
```
xiaoxiang781216, line 34: It's the caller's responsibility to not call metal_finish in case of metal_init return fail.

arnopo:
alternative proposed in https://github.com/OpenAMP/libmetal/pull/239

arnopo:
@edmooring I close this PR as https://github.com/OpenAMP/libmetal/pull/239 should fix the issue. don't hesitate to reopen it if needed

PR #188, opened 2021-11-13
Add support for registering interrupts on Zephyr

twpedersen:
This PR simply wires metal interrupt APIs to the Zephyr APIs.

### DIFF
--- a/lib/system/zephyr/irq.c
+++ b/lib/system/zephyr/irq.c
@@ -10,6 +10,7 @@
  */
 
 #include <metal/errno.h>
+#include <metal/irq_controller.h>
 #include <metal/irq.h>
 #include <metal/sys.h>
 #include <metal/log.h>
@@ -17,6 +18,12 @@
 #include <metal/list.h>
 #include <metal/utilities.h>
 #include <metal/alloc.h>
+
+#ifdef CONFIG_DYNAMIC_INTERRUPTS
+#define MAX_IRQS CONFIG_NUM_IRQS
+
+static struct metal_irq irqs[MAX_IRQS];
+#endif /* CONFIG_DYNAMIC_INTERRUPTS */
 
 unsigned int metal_irq_save_disable(void)
 {
@@ -28,3 +35,60 @@ void metal_irq_restore_enable(unsigned int flags)
 	irq_unlock(flags);
 }
 
+#ifdef CONFIG_DYNAMIC_INTERRUPTS
+static void metal_zephyr_irq_set_enable(struct metal_irq_controller *irq_cntr,
+				     int irq, unsigned int state)
+{
+	if (irq < irq_cntr->irq_base ||
+	    irq >= irq_cntr->irq_base + irq_cntr->irq_num) {
+		metal_log(METAL_LOG_ERROR,
+			  "%s: invalid irq %d\n", __func__, irq);
+		return;
+	} else if (state == METAL_IRQ_ENABLE) {
+		irq_enable(irq);
+	} else {
+		irq_disable(irq);
+	}
+}
+
+static void metal_zephyr_irq_isr(const void *arg)
+{
+	unsigned int vector = (unsigned int) arg;
+
+	if (vector >= MAX_IRQS) {
+		return;
+	}
+	(void) metal_irq_handle(&irqs[vector], (int)vector);
+}
+
+static int metal_zephyr_irq_register(struct metal_irq_controller *cntr, int irq,
+				     metal_irq_handler hd, void *arg)
+{
+	irqs[irq].hd = hd;
+	irqs[irq].arg = arg;
+	irq_connect_dynamic(irq, 0, metal_zephyr_irq_isr, (void *)irq, 0);
+	return 0;
+}
+
+static METAL_IRQ_CONTROLLER_DECLARE(zephyr_irq_cntr,
+				    0, MAX_IRQS,
+				    NULL,
+				    metal_zephyr_irq_set_enable,
+				    metal_zephyr_irq_register,
+				    irqs);
+#endif	/* CONFIG_DYNAMIC_INTERRUPTS */
+
+int metal_zephyr_irq_init(void)
+{
+#ifdef CONFIG_DYNAMIC_INTERRUPTS
+	int ret;
+
+	ret =  metal_irq_register_controller(&zephyr_irq_cntr);
+	if (ret < 0) {
+		metal_log(METAL_LOG_ERROR,
+			  "%s: register irq controller failed.\n", __func__);
+		return ret;
+	}
+#endif	/* CONFIG_DYNAMIC_INTERRUPTS */
+	return 0;
+}
--- a/lib/system/zephyr/sys.h
+++ b/lib/system/zephyr/sys.h
@@ -41,6 +41,17 @@ struct metal_state {
 	struct metal_common_state common;
 };
 
+/**
+ * @brief	metal_zephyr_irq_init
+ *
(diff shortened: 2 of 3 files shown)


arnopo:
Hello @twpedersen 

Could you details the context 
- what is /are interrupts relying on this mechanism?
- Why do you need to handle interput in the lib metal instead of implementing directly the management in Zephyr project/drivers?

lib/system/zephyr/irq.c
```diff
@@ -18,6 +19,12 @@
 #include <metal/utilities.h>
 #include <metal/alloc.h>
 
+#ifdef CONFIG_DYNAMIC_INTERRUPTS
+#define MAX_IRQS CONFIG_NUM_IRQS
```
carlocaione, line 23: You should also take into account `CONFIG_GEN_IRQ_START_VECTOR`

lib/system/zephyr/irq.c
```diff
@@ -28,3 +35,60 @@ void metal_irq_restore_enable(unsigned int flags)
...
+	}
+	(void) metal_irq_handle(&irqs[vector], (int)vector);
+}
+
+static int metal_zephyr_irq_register(struct metal_irq_controller *cntr, int irq,
+				     metal_irq_handler hd, void *arg)
+{
+	irqs[irq].hd = hd;
+	irqs[irq].arg = arg;
+	irq_connect_dynamic(irq, 0, metal_zephyr_irq_isr, (void *)irq, 0);
```
carlocaione, line 69: same here, `irq` cannot be `< CONFIG_GEN_IRQ_START_VECTOR`

lib/system/zephyr/sys.h
```diff
@@ -41,6 +41,17 @@ struct metal_state {
...
 };
 
+/**
+ * @brief	metal_zephyr_irq_init
+ *
+ * Initialize the metal IRQ controller abstraction.
+ *
+ * @return 0 for success, or negative value for failure.
+ *
+ * @return 0 for success, or negative value for failure
```
carlocaione, line 51: duplicate comment

twpedersen:
Hi @arnopo 

> Hello @twpedersen
> 
> Could you details the context
> 
>     * what is /are interrupts relying on this mechanism?
> 
>     * Why do you need to handle interput in the lib metal instead of implementing directly the management in Zephyr project/drivers?

Indeed. I was working on porting the zephyr [openamp sample](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/ipc/openamp_rsc_table) to a new platform and needed a way to register/handle interrupts. As you can see the sample already calls into libmetal, so to remain consistent we added support for interrupt handling.

Later it did occur to me that abstracting away zephyr from within a zephyr application by calling libmetal which is just a wrapper around zephyr APIs is kind of pointless :)

Can you think of a reason the zephyr example calls into libmetal at all?

If the intent of libmetal is to let an application easily retarget the OS or hardware, then maybe this PR is still useful.

arnopo:
> 



> Hi @arnopo
> 
> > Hello @twpedersen
> > Could you details the context
> > ```
> > * what is /are interrupts relying on this mechanism?
> > 
> > * Why do you need to handle interput in the lib metal instead of implementing directly the management in Zephyr project/drivers?
> > ```
> 
> Indeed. I was working on porting the zephyr [openamp sample](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/ipc/openamp_rsc_table) to a new platform and needed a way to register/handle interrupts. As you can see the sample already calls into libmetal, so to remain consistent we added support for interrupt handling.

you need an interrupt or a mailbox(IPM)  for the Inter-processor communication?

> 
> Later it did occur to me that abstracting away zephyr from within a zephyr application by calling libmetal which is just a wrapper around zephyr APIs is kind of pointless :)
> 
> Can you think of a reason the zephyr example calls into libmetal at all?
> 
> If the intent of libmetal is to let an application easily retarget the OS or hardware, then maybe this PR is still useful.

For this, the application should rely on libmetal for all interrupts... Seems to me not reliable, but confusing.
So I would prefer to limit the use of libmetal to porting the open-amp to different hardware and OS, and not address the issue of application portability.

In other words, I would prefer to wait for a concrete need to integrate this work.

@carlocaione, @edmooring:
Any opinion on this?

carlocaione:
> As you can see the sample already calls into libmetal, so to remain consistent we added support for interrupt handling.

You do not need to directly support interrupt handling, the IPM / MBOX driver is taking care of that for you.

> Can you think of a reason the zephyr example calls into libmetal at all?

As @arnopo said (in the context of OpenAMP) `libmetal` is a nice way to make it easy to port OpenAMP to different OSes and abstract it from the underlying HW. For example [rproc_virtio_init_vring()](https://github.com/OpenAMP/open-amp/blob/master/lib/include/openamp/remoteproc_virtio.h#L91-L94) takes a `metal_io_region *` as parameter, so it is handy to have the application calling into libmetal to get that when using OpenAMP.

> In other words, I would prefer to wait for a concrete need to integrate this work.

IMO if we consider `libmetal` as an auxiliary library to OpenAMP, this PR is not useful. It could be useful if we would consider `libmetal` as a generic abstraction library also for standalone applications (that maybe need to use IRQs).

twpedersen:
OK so libmetal is simply meant to help port OpenAMP to new platforms. In that context maybe this PR doesn't make much sense.

I was porting the Zephyr `openamp_rsc_table` to the Xilinx zynq platform which doesn't have IPM, but relies on IPI (software interrupts) to signal on the vrings.

Thank you for your review.

carlocaione:
If you are willing to, I think that having a new IPC service backend using a resource table could be very welcome in upstream zephyr (instead of simply porting the sample).

twpedersen:
> If you are willing to, I think that having a new IPC service backend using a resource table could be very welcome in upstream zephyr (instead of simply porting the sample).

OK, I will take a look.

twpedersen:
@carlocaione just to confirm I'm on the right track: you're suggesting adding an `subsys/ipc/ipc_service/backends/ipc_rpmsg_rsc_table.c` ?

Regarding the lack of IPM on zynq: `ipc_rpmsg_static_vrings.c` uses `mbox` to trigger the vrings. Adding a dummy `mbox` driver for the zynq platform which internally just registers a softirq handler for the recv callback and triggers softirq on send could work, but of course the data pointer would be ignored. This would be a huge hack.

Maybe it makes sense to add switchable (at build time) IPI implementations (`mbox`/`ipm`/`sgi`) in `subsys/ipc/ipc_service/lib/` or so?

carlocaione:
> @carlocaione just to confirm I'm on the right track: you're suggesting adding an `subsys/ipc/ipc_service/backends/ipc_rpmsg_rsc_table.c` ?

Correct.

> Regarding the lack of IPM on zynq: `ipc_rpmsg_static_vrings.c` uses `mbox` to trigger the vrings. Adding a dummy `mbox` driver for the zynq platform which internally just registers a softirq handler for the recv callback and triggers softirq on send could work, but of course the data pointer would be ignored. This would be a huge hack.

It's not an hack. The MBOX API explicitly supports the signalling mode, see https://github.com/zephyrproject-rtos/zephyr/blob/main/include/drivers/mbox.h#L56-L59

> Maybe it makes sense to add switchable (at build time) IPI implementations (`mbox`/`ipm`/`sgi`) in `subsys/ipc/ipc_service/lib/` or so?

It makes sense. The only problem I see is that the MBOX API was introduced because the IPM API does not support multiple channels, so if you use IPM there can be only one single user of the instance.

PR #190, opened 2021-11-22
lib: fix compilation error with gcc++

arnopo:
The header files including must be outside of the extern "C" { statement.
To be able to be included from C++ code.

Else such error can occur (reproduce on STM32MP1Cube distribution)
gcc/arm-none-eabi/include/c++/10.2.1/atomic:1467:3: error: template with C linkage
 1467 |   template<typename _ITp>
      |   ^~~~~~~~
In file included from STM32CubeMP1/Middlewares/Third_Party/OpenAMP/open-amp/lib/include/openamp/rpmsg.h:16,
                 from STM32CubeMP1/Middlewares/Third_Party/OpenAMP/open-amp/lib/include/openamp/open_amp.h:11,
                 from empty_main.cpp:1:
STM32CubeMP1/Middlewares/Third_Party/OpenAMP/libmetal/lib/include/metal/mutex.h:16:1: note: 'extern "C"' linkage
started here
   16 | extern "C" {
      | ^~~~~~~~~~

### DIFF
--- a/lib/alloc.h
+++ b/lib/alloc.h
@@ -36,12 +36,12 @@ static inline void *metal_allocate_memory(unsigned int size);
  */
 static inline void metal_free_memory(void *ptr);
 
-#include <metal/system/@PROJECT_SYSTEM@/alloc.h>
-
 /** @} */
 
 #ifdef __cplusplus
 }
 #endif
+
+#include <metal/system/@PROJECT_SYSTEM@/alloc.h>
 
 #endif /* __METAL_ALLOC__H__ */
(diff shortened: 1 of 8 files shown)


edmooring (approved):
Looks good to me. Do we have any C++ test cases in our CI?

arnopo:
> Looks good to me. Do we have any C++ test cases in our CI?

Unfortunately not. But we can start by creating an issue to track it.

I use the following project to test the API:
https://github.com/sergeyblagodarnyy/rpmsg_cpp_compatibility. 

But it does not provide exhaustive tests

PR #198, opened 2022-01-17
freertos: io: fix IAR compile error due to 'void *' variable

arnopo:
Fix EWARM compilation error due to void pointer usage.
This fix has already been implemented for the generic system, but
was missing for the FreeRTOS.

Signed-off-by: Arnaud Pouliquen <arnaud.pouliquen@st.com>


edmooring (approved):
Looks good to go.

PR #200, opened 2022-02-01
lib: system: zephyr: implement __metal_sleep_usec

emfend:
The currently missing implementation causes invalid timeouts when sending
remoteproc messages over virtIO. With this change, the specified timeout
of 15 seconds can be achieved. This seems to be the only usage in Zephyr.

Signed-off-by: Matthias Fend <matthias.fend@emfend.at>


edmooring (approved):
Not a Zephyr expert, but looks good to me.

PR #203, opened 2022-04-06
Doc: fix doc generation warnings in system sleep.h files

arnopo:
Fix "unbalanced grouping commands" warnings by cleaning up
the end-of-block occurrences, defined without starting tag.
No need to define an API interface block in these files, the
interface is already described in the main sleep.h file.


edmooring (approved):
Looks good to me.

PR #204, opened 2022-04-13
Create an eventfd object with `EFD_CLOEXEC` flag, to avoid leaking file descriptor to child processes

Tosainu:
The file descriptor `irq_notify_fd` does not have the close-on-exec (`FD_CLOEXEC`) flag so it possibly leaks to child processes.

This pull request adds an `EFD_CLOEXEC` flag to the second argument of the `exentfd(2)` call, which tells it to set an `FD_CLOEXEC` flag on the new file descriptor.


arnopo:
Hello @Tosainu 
Could please fix the failed check? you don' follow the requested commit message format,  by clicking "details" it you will have details on the issues.

Tosainu:
Hello. Thanks for your comments. I fixed failed compliance checks.

edmooring (approved):
This change looks good to me. Since the usage of this fd is largely internal to libmetal, leaving it open on exec() will leak a file descriptor.

PR #208, opened 2022-04-26
Simplify zephyr MACHINE and map processor arm64 -> aarch64

quic-egmc:
Replace the architecture-specific machine directories with a common
C function for polling which calls to arch_cpu_idle.

Add a special case to map Zephyr arm64 to libmetal aarch64 in defining
PROJECT_PROCESSOR variable.

Set the MACHINE variable based CONFIG_ARCH instead of testing
individual arch config flags.


carlocaione (approved):

edmooring (approved):
Looks good to go.

quic-egmc:
Thanks - I'm still doing some testing under arm and arm64

lib/system/zephyr/poll.c
```diff
@@ -1,22 +1,23 @@
...
- * Copyright (c) 2021 Carlo Caione <ccaione@baylibre.com>
+ * Copyright (c) 2022 Qualcomm Innovation Center, Inc.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
 /*
- * @file	zephyr/riscv/sys.c
- * @brief	machine specific system primitives implementation.
+ * @file	zephyr/poll.c
```
arnopo, line 8: Could you keep sys.c to  harmonize the file naming with other platforms.

arnopo (changes requested):
LGTM, could you just rename `pool.c `to `sys.c` that's `metal_generic_default_poll` is define in sys.c for all platforms?
thanks

arnopo:
> Thanks - I'm still doing some testing under arm and arm64
Please let me know when your tests are finished (maybe it is already finished).

lib/system/zephyr/poll.c
quic-egmc: done.  force-pushed changes to rename poll.c to sys.c and update CMakeLists.txt accordingly.

(I can't believe that in 2022 github still can't track changes across force-pushes - other tools like gitlab have been doing this for years)

arnopo: The delta is visible by clicking on the compare button or the force_push ;-)

arnopo (approved):
LGTM

lib/system/zephyr/poll.c
quic-egmc: oh, thanks!  somehow I expected to see this under Files changed but it appears it's only visible in the PR "conversation".  that's what I get for hanging out in gitlab for a while!

PR #210, opened 2022-05-10
zephyr: update include paths to use <zephyr/...>

gmarull:
Zephyr has prefixed all of its includes with <zephyr/...>. While the old
mode can still be used (CONFIG_LEGACY_INCLUDE_PATH) and is still enabled
by default, it's better to be prepared for its removal in the future.


gmarull:
cc: @carlocaione

carlocaione (approved):

gmarull:
@arnopo the first adds `<zephyr/>` prefix, the second changes `<zephyr/zephyr.h>` to `<zephyr/kernel.h>`.

edmooring (approved):
Looks good to me.

PR #213, opened 2022-05-25
Virtio exp

danmilea:
virtio-mmio framework for libmetal / open-amp


lib/system/zephyr/generic/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+collect (PROJECT_LIB_HEADERS sys.h)
```
arnopo, line 1: https://github.com/OpenAMP/libmetal/pull/208 probably makes this patch obsolete
I can move the` virtio_exp` branch on top of it

lib/system/zephyr/sys.h
```diff
@@ -31,7 +31,7 @@ extern "C" {
 }
 
 #ifndef METAL_MAX_DEVICE_REGIONS
-#define METAL_MAX_DEVICE_REGIONS 1
+#define METAL_MAX_DEVICE_REGIONS 2
```
arnopo, line 34: This should be defined in zephyr. in prj.conf or in Cmakelists.txt
for instance: https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/subsys/ipc/openamp_rsc_table/CMakeLists.txt#L15

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+#
```
arnopo, line 1: This one is OK for the virtio_exp branch to avoid to fork the zephyrproject-rtoss/open-amp and zephyrproject-rtos/libmetal repos.
 Please, could you add the information in commit subject

lib/processor/arm64
```diff
@@ -0,0 +1 @@
+aarch64
```
arnopo, line 1: should also be addressed in https://github.com/OpenAMP/libmetal/pull/208

arnopo (commented):
Few comments. could you have a look to https://github.com/OpenAMP/libmetal/pull/208, and check if it could replace 2 of your commits?
i can also update the virtio_exp branch to point on top of the https://github.com/OpenAMP/libmetal/pull/208 commits

PR #217, opened 2022-07-26
New routine to find a node on a list

tammyleino:
metal_list_find_node is used to ensure a node is not on a list before adding it
Signed-off-by: Tammy Leino <tammy_leino@mentor.com>

### DIFF
--- a/lib/list.h
+++ b/lib/list.h
@@ -12,6 +12,7 @@
 #ifndef __METAL_LIST__H__
 #define __METAL_LIST__H__
 
+#include <stdbool.h>
 #include <stdlib.h>
 
 #ifdef __cplusplus
@@ -96,6 +97,18 @@ static inline struct metal_list *metal_list_first(struct metal_list *list)
 	for ((node) = (list)->next;		\
 	     (node) != (list);			\
 	     (node) = (node)->next)
+
+static inline bool metal_list_find_node(struct metal_list *list,
+					struct metal_list *node)
+{
+	struct metal_list *n;
+
+	metal_list_for_each(list, n) {
+		if (n == node)
+			return true;
+	}
+	return false;
+}
 /** @} */
 
 #ifdef __cplusplus


lib/list.h
```diff
@@ -96,6 +96,18 @@ static inline struct metal_list *metal_list_first(struct metal_list *list)
 	for ((node) = (list)->next;		\
 	     (node) != (list);			\
 	     (node) = (node)->next)
+
+static inline int metal_list_find_node(struct metal_list *list,
```
arnopo, line 100: return a boolean?

lib/list.h
```diff
@@ -96,6 +96,18 @@ static inline struct metal_list *metal_list_first(struct metal_list *list)
...
+				       struct metal_list *node)
+{
+	struct metal_list *n;
+
+	metal_list_for_each(list, n) {
+		if (n == node)
+			break;
```
arnopo, line 107: return true;

lib/list.h
```diff
@@ -96,6 +96,18 @@ static inline struct metal_list *metal_list_first(struct metal_list *list)
...
+	}
+	return (n == list) ? 0 : 1;
```
arnopo, line 109: return false;

arnopo (commented):
Seems that the checkpatch fail is a false positive. we can ignore it.
Minors comment.

tammyleino:
@arnopo  I addressed your comments but am seeing the checkpatch failure still.

arnopo (approved):

arnopo:
> @arnopo I addressed your comments but am seeing the checkpatch failure still.

It is a false positif check patch. You can ignore it.

lib/list.h
```diff
@@ -13,6 +13,7 @@
 #define __METAL_LIST__H__
 
 #include <stdlib.h>
+#include <stdbool.h>
```
arnopo, line 16: nitpicking: please declare it in an alphabetical order so before `#include <stdlib.h>`

edmooring (approved):
Looks good to me.

PR #218, opened 2022-08-27
zynqmp_r5: replace Xil_SetMPURegion with Xil_MemMap

kernelchuk:
The metal_machine_io_mem_map() used to call Xil_SetMPURegion directly
and could get into an infinite loop. Replace the faulty code in
metal_machine_io_mem_map() with Xil_MemMap() call and assert it succeeds.

Signed-off-by: Sergei Korneichuk <sergei.korneichuk@amd.com>
---
fixes: 51e08c67556474762257a825fc92cf8dbd302388

### DIFF
--- a/lib/system/freertos/zynqmp_r5/sys.c
+++ b/lib/system/freertos/zynqmp_r5/sys.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2016, Xilinx Inc. and Contributors. All rights reserved.
+ * Copyright (C) 2022, Advanced Micro Devices, Inc.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -19,8 +20,6 @@
 #include "xil_mpu.h"
 #include "xreg_cortexr5.h"
 #include "xscugic.h"
-
-#define MPU_REGION_SIZE_MIN 0x20
 
 void sys_irq_restore_enable(unsigned int flags)
 {
@@ -61,27 +60,16 @@ void metal_weak metal_generic_default_poll(void)
 	metal_asm volatile("wfi");
 }
 
+/**
+ * The code moved to cortexr5/xil_mpu.c:Xil_MemMap()
+ * NULL in pa masks possible Xil_MemMap() errors.
+ */
 void *metal_machine_io_mem_map(void *va, metal_phys_addr_t pa,
 			       size_t size, unsigned int flags)
 {
-	size_t rsize = MPU_REGION_SIZE_MIN;
-	metal_phys_addr_t base_pa;
+	void *__attribute__((unused)) physaddr;
 
-	if (!flags)
-		return va;
-	while (1) {
-		if (rsize < size) {
-			rsize <<= 1;
-			continue;
-		} else {
-			base_pa = pa & ~(rsize - 1);
-			if ((base_pa + rsize) < (pa + size)) {
-				rsize <<= 1;
-				continue;
-			}
-			break;
-		}
-	}
-	Xil_SetMPURegion(base_pa, rsize, flags);
+	physaddr = Xil_MemMap(pa, size, flags);
+	metal_assert(physaddr == (void *)pa);
 	return va;
 }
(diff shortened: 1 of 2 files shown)


TanmayShah-xilinx:
@arnopo, @edmooring 

```
#51: FILE: lib/system/freertos/zynqmp_r5/sys.c:69:
[23](https://github.com/OpenAMP/libmetal/runs/8046721200?check_suite_focus=true#step:6:24)
+	void *__attribute__((unused)) physaddr = Xil_MemMap(pa, size, flags);
[24](https://github.com/OpenAMP/libmetal/runs/8046721200?check_suite_focus=true#step:6:25)
```

Above warning is false positive I believe. As it is not function definition, but Xil_MemMap is function use.
However, would you recommend not to use __attribute__() in the same line where variable is initialized?
I don't see any downside of it, but I may be wrong.

arnopo:
> @arnopo, @edmooring
> 
> ```
> #51: FILE: lib/system/freertos/zynqmp_r5/sys.c:69:
> [23](https://github.com/OpenAMP/libmetal/runs/8046721200?check_suite_focus=true#step:6:24)
> +	void *__attribute__((unused)) physaddr = Xil_MemMap(pa, size, flags);
> [24](https://github.com/OpenAMP/libmetal/runs/8046721200?check_suite_focus=true#step:6:25)
> ```
> 
> Above warning is false positive I believe. As it is not function definition, but Xil_MemMap is function use. However, would you recommend not to use **attribute**() in the same line where variable is initialized? I don't see any downside of it, but I may be wrong.

Following code seeems not generate this  false positive:
```
void *metal_machine_io_mem_map(void *va, metal_phys_addr_t pa,
			       size_t size, unsigned int flags)
{
	void *__attribute__((unused)) physaddr;

	physaddr = Xil_MemMap(pa, size, flags);
	metal_assert(physaddr == (void *)pa);
	return va;
}
```

kernelchuk:
Thank you Arnaud, I fixed the code, I should have done it when it was flagged before sending the first PR. Not sure why Zephyr build fails. Any hints ?

arnopo:
> Thank you Arnaud, I fixed the code, I should have done it when it was flagged before sending the first PR. Not sure why Zephyr build fails. Any hints ?

No idea yet, probably a problem in the IC.  I tried to rerun the action without success. I need to look deeper. Just ignore the check for now.

arnopo:
@bentheredonethat
Any feedback on this PR?
Thanks

bentheredonethat:
@arnopo  i am in support of this patch. AMD BSP provides this so we should leverage as proposed by Sergei.

edmooring (commented):
Looks good to me.

edmooring (approved):

arnopo:
Please could you rebase on main to rerun the Ci? CI issue should be fixed in https://github.com/OpenAMP/libmetal/pull/219

kernelchuk:
Hi Arnaud, Ed,
I have about 30 commits with bug fixes in our AMD/Xilinx internal repos and I'd like to upstream them.
I can
  1.  Create PRs for each of them
  2.  Combine all or some of them in a one PR
What is your preferred method of up-steaming these commits?

lib/system/freertos/zynqmp_r5/sys.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, Xilinx Inc. and Contributors. All rights reserved.
+ * Copyright (c) 2016 - 2022, Xilinx Inc. and Contributors. All rights reserved.
```
arnopo, line 2: nitpicking : just replace 2016 by 2022

lib/system/generic/zynqmp_r5/sys.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, Xilinx Inc. and Contributors. All rights reserved.
+ * Copyright (c) 2016 - 2022, Xilinx Inc. and Contributors. All rights reserved.
```
arnopo, line 2: same here

arnopo (changes requested):
just an update for the copyright else LGTM

edmooring:
Agreed on the 30 commits. As a rough guideline, combining commits that don't overlap and only affect one file each is fine. A multi-file commit probably deserves its own PR.

arnopo (approved):

PR #220, opened 2022-09-16
cmake: set PROJECT_EC_FLAGS to be GLOBAL property

kernelchuk:
Declare PROJECT_EC_FLAGS to be a global property
to simplify code for target_compile_options.

Signed-off-by: Sergei Korneichuk <sergei.korneichuk@amd.com>


edmooring (approved):
Looks good to go.

arnopo (approved):

PR #222, opened 2022-10-04
API to validate address within io block

tammyleino:
Ensures the specified address falls within the valid range of the io block
Signed-off-by: Tammy Leino <tammy_leino@mentor.com>

### DIFF
--- a/lib/io.c
+++ b/lib/io.c
@@ -139,3 +139,15 @@ int metal_io_block_set(struct metal_io_region *io, unsigned long offset,
 	return retlen;
 }
 
+bool metal_io_is_block_valid(struct metal_io_region *io, unsigned long offset,
+			     const void *addr, int len)
+{
+	if (!io || !addr)
+		return false;
+
+	if ((addr >= (io->virt + offset)) && ((addr + len) < (io->virt + io->size)))
+		return true;
+
+	return false;
+}
+
--- a/lib/io.h
+++ b/lib/io.h
@@ -363,6 +363,17 @@ int metal_io_block_write(struct metal_io_region *io, unsigned long offset,
 int metal_io_block_set(struct metal_io_region *io, unsigned long offset,
 		       unsigned char value, int len);
 
+/**
+ * @brief	validate an address within a block.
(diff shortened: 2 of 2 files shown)


edmooring (approved):
This looks good to me. There may be a case where a false positive can happen (addr + len overflows), but that shouldn't stop this improvement from going in.

lib/io.c
```diff
@@ -139,3 +139,15 @@ int metal_io_block_set(struct metal_io_region *io, unsigned long offset,
 	return retlen;
 }
 
+bool metal_io_is_block_valid(struct metal_io_region *io, unsigned long offset,
+			     const void *addr, int len)
+{
+	if (!io || !addr)
+		return false;
+
+	if ((addr >= (io->virt + offset)) && ((addr + len) < (io->virt + io->size)))
```
arnopo, line 148: I missed that, you increment void pointers. you have to cast them before.
Could you also add a header to describe the parameters?
As this commit need update could you also add the offerflow test pointed out by @Ed ?

arnopo (changes requested):

tammyleino:
> This looks good to me. There may be a case where a false positive can happen (addr + len overflows), but that shouldn't stop this improvement from going in.

@edmooring  Thank you very much for your eyes on this.  Can you please provide an example of the overflow so I am clear?

lib/io.c
tammyleino: @arnopo Yes, no problem - thank you for the suggestion.  Do you prefer uint8_t* or unsigned char*

I see virt cast to uint8_t in io.h but represented as unsigned char* in io.c

edmooring:
@tammyleino For the overflow, consider the following on a 32-bit architecture:
io->virt = 0xffffff00
io->size = 0x80
addr = 0xffffff08
len = 0x100
Unless I have missed something, addr + len will overflow to 0x8 which will pass the test of (addr + len < io->virt + io->size) but still access memory

lib/io.c
arnopo: uint8_t seems more reliable  but both are ok

PR #231, opened 2023-03-31
ZyncMP Inter-Processor Interrupt / Interface API

kernelchuk:
Hi, I'm trying to review and upstream our Inter-Processor Interrupt / Interface (IPI) code reorganization.
I think O_WRONLY should not be flagged as a spelling error.


arnopo:
I took a look at your PR.
it is only example of implementation, right?

We have now  the [system reference](https://github.com/OpenAMP/openamp-system-reference) repository
which  would be (in my opinion) the good places for examples. 

I would prefer to move this to the system_reference instead of overloading the library with examples related to one platform.

@edmooring , @wmamills  any opinion on That?

edmooring:
> I took a look at your PR. it is only example of implementation, right?
> 
> We have now the [system reference](https://github.com/OpenAMP/openamp-system-reference) repository which would be (in my opinion) the good places for examples.
> 
> I would prefer to move this to the system_reference instead of overloading the library with examples related to one platform.
> 
> @edmooring , @wmamills any opinion on That?

@arnopo I agree that the examples would mostly be better in open amp-system-reference. I don't think that this is an issue for this upcoming release.

edmooring (approved):
This looks good to me, but should be moved to the system reference repo. I think much of this code is not Xilinx-specific and could be moved to a more general build.

kernelchuk:
Hi Ed and Arnaud,
Thank you for your review. I agree that we should move our libmetal zynqmp_amp_demo examples to [system reference](https://github.com/OpenAMP/openamp-system-reference) repository.  The changes are to nine (9) files out of 13 directories, 48 files under the examples directory.

Our AMD/Xilinx internal libmetal branch is about 20 commits ahead of upstream and I’m in the process of moving our development to “upstream-first” mode. I think it’s easier to move to the new repo after I upstream the delta. This will give me time to adjust our Yocto build and our internal CI scripts to use the new repo.

wmamills:
I can't see why this was closed.  I reopened it.  If there was a reason it was close please supply it.

kernelchuk:
Closing due to https://github.com/OpenAMP/openamp-system-reference/pull/21

PR #236, opened 2023-05-30
Wrap vars ifndef

bentheredonethat:
Enable demos to be configured via CMake configure or toolchain file.

This enables the demos to be configured for Versal and Versal_Net platforms.


bentheredonethat:
Hi @arnopo  @edmooring 

the error here does not seem to apply to the patches given. please review the set when able.

this makes it so we can configure the apps via cmake and thus the demos more scalable.

i understand there is a pending move to system-reference repo for apps. since the apps have not yet been moved, can we review and merge here first and then at the time of the move we can then begin to send patches to the new repo?

thanks

TanmayShah-xilinx (approved):

kernelchuk:
Ben, Tanmay, 

This PR contradicts two decisions: 
1. Our plan to use a (re-)generated file for IPI_*, SHM_*, TTC_*, etc. values.
2. This https://github.com/OpenAMP/libmetal/pull/231#issuecomment-1499325357 about changes in the example code.

bentheredonethat:
closing for now

PR #239, opened 2023-06-04
Add reference count to metal_common_state

xiaoxiang781216:
to allow to invoke metal_init and metal_finish multiple times

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -11,24 +11,30 @@ int metal_init(const struct metal_init_params *params)
 {
 	int error = 0;
 
-	memset(&_metal, 0, sizeof(_metal));
+	if (atomic_fetch_add(&_metal.common.ref_count, 1) == 0) {
+		memset(&_metal.common + 1, 0,
+		       sizeof(_metal) - sizeof(_metal.common));
 
-	_metal.common.log_handler   = params->log_handler;
-	_metal.common.log_level     = params->log_level;
+		_metal.common.log_handler = params->log_handler;
+		_metal.common.log_level   = params->log_level;
 
-	metal_list_init(&_metal.common.bus_list);
-	metal_list_init(&_metal.common.generic_shmem_list);
-	metal_list_init(&_metal.common.generic_device_list);
+		metal_list_init(&_metal.common.bus_list);
+		metal_list_init(&_metal.common.generic_shmem_list);
+		metal_list_init(&_metal.common.generic_device_list);
 
-	error = metal_sys_init(params);
-	if (error)
-		return error;
+		error = metal_sys_init(params);
+		if (error)
+			atomic_fetch_sub(&_metal.common.ref_count, 1);
+	}
 
 	return error;
 }
 
 void metal_finish(void)
 {
-	metal_sys_finish();
-	memset(&_metal, 0, sizeof(_metal));
+	if (atomic_fetch_sub(&_metal.common.ref_count, 1) == 1) {
+		metal_sys_finish();
+		memset(&_metal.common + 1, 0,
+		       sizeof(_metal) - sizeof(_metal.common));
+	}
 }
--- a/lib/sys.h
+++ b/lib/sys.h
@@ -17,6 +17,7 @@
 
 #include <metal/log.h>
 #include <metal/list.h>
+#include <metal/atomic.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -61,6 +62,9 @@ struct metal_init_params {
  * specific singleton data structure (@see _metal).
  */
 struct metal_common_state {
+	/** reference count to track metal_init/metal_finish. */
+	atomic_int			ref_count;
+
 	/** Current log level. */
 	enum metal_log_level		log_level;
 


xiaoxiang781216:
@arnopo is Zephyr CI broken?

lib/init.c
```diff
@@ -11,24 +11,26 @@ int metal_init(const struct metal_init_params *params)
...
 
 	return error;
 }
 
 void metal_finish(void)
 {
-	metal_sys_finish();
-	memset(&_metal, 0, sizeof(_metal));
+	if (atomic_fetch_sub(&_metal.common.ref_count, 1) == 1)
+		metal_sys_finish();
```
arnopo, line 35: The `metal_state _metal` depends on the architecture, I would prefer to keep the memset to avoid side effect.

lib/init.c
```diff
@@ -11,24 +11,26 @@ int metal_init(const struct metal_init_params *params)
...
-	metal_list_init(&_metal.common.generic_shmem_list);
-	metal_list_init(&_metal.common.generic_device_list);
-
-	error = metal_sys_init(params);
-	if (error)
-		return error;
+		error = metal_sys_init(params);
+		if (error) {
+			atomic_fetch_sub(&_metal.common.ref_count, 1);
+			return error;
```
arnopo, line 25: The return was and is still useless . could be suppressed

arnopo (changes requested):
LGTM after adressing comments

lib/init.c
xiaoxiang781216: Do you want to change the return value of metal_init and metal_sys_init to void or just ignore the value returned from metal_sys_init?

arnopo: unclear..., it was just to suppress the "return error;" duplication
My proposal was to have something like this
```
int metal_init(const struct metal_init_params *params)
{
	int error = 0;
	if (atomic_fetch_add(&_metal.common.ref_count, 1) == 0) {

[...]
		error = metal_sys_init(params);
		if (error) {
			atomic_fetch_sub(&_metal.common.ref_count, 1);
}

	return error;
```

xiaoxiang781216: Done.

arnopo (approved):

arnopo:
@edmooring, @TanmayShah-xilinx 
Could you review and confirm that fix the issue related to the PR  https://github.com/OpenAMP/libmetal/pull/185 
Thanks!

arnopo:
> @edmooring, @TanmayShah-xilinx Could you review and confirm that fix the issue related to the PR #185 Thanks!

Gentle reminder

arnopo:
@edmooring-xilinx and @TanmayShah-xilinx:  as no comment I consider that the PR is good to be merged.

TanmayShah-xilinx:
Hi Arnaud,

I missed this review request. I will pull I changes and test it. I will let you know if I face any issues.
Meanwhile the patch looks good to me as well and it’s ok to merge. I don’t see any obvious issues.

Thanks,
Tanmay

From: Arnaud Pouliquen ***@***.***>
Date: Monday, July 3, 2023 at 2:12 AM
To: OpenAMP/libmetal ***@***.***>
Cc: Shah, Tanmay ***@***.***>, Mention ***@***.***>
Subject: Re: [OpenAMP/libmetal] Add reference count to metal_common_state (PR #239)

@edmooring-xilinx<https://github.com/edmooring-xilinx> and @TanmayShah-xilinx<https://github.com/TanmayShah-xilinx>: as no comment I consider that the PR is good to be merged.

arnopo:
FYI this commit introduce regression in Zephyr, in openamp CI when run following command:
`west build --sysbuild -b lpcxpresso54114_m4 samples/subsys/ipc/openamp/`

Issue on link :
`./zephyrproject/modules/hal/libmetal/libmetal/lib/init.c:35: undefined reference to '__atomic_fetch_sub_4'`

Linked to the stdatomic lib... I will investigate it.

TanmayShah-xilinx:
Hi @arnopo,

I looked into this issue. linking to atomic library is missing. Usually we can resolve with -latomic option, but here runtime linking won't work.

I am trying to find out how to include atomic library statically in libmetal.

Any other possible solution? I also see GCC should provide in-built atomic APIS, but I believe cortex-m4 atomic operations isn't supported with GCC yet.

So, including atomic library statically is our only option looks like.

arnopo:
@TanmayShah-xilinx:
Thanks for the feedback I came to the same conclusion.
The atomic management depends on OS, compiler but also processor. It just a nightmare to manage it in libmetal....
For Zephyr , a specific API is used  (https://elixir.bootlin.com/zephyr/latest/source/include/zephyr/sys/atomic.h)
The CI fails for a Cortex-M0.

@xiaoxiang781216 :
Regarding regression introduce I plan to remove this commit from the main branch.
We have to rework the solution,  if this fix is needed. 
Do you need atomic to manage the init?

arnopo:
I rebased the main branch to remove it

xiaoxiang781216:
> @TanmayShah-xilinx: Thanks for the feedback I came to the same conclusion. The atomic management depends on OS, compiler but also processor. It just a nightmare to manage it in libmetal.... For Zephyr , a specific API is used (https://elixir.bootlin.com/zephyr/latest/source/include/zephyr/sys/atomic.h) The CI fails for a Cortex-M0.
> 

I use atomic API because it is abstracted in libmetal.

> @xiaoxiang781216 : Regarding regression introduce I plan to remove this commit from the main branch. We have to rework the solution, if this fix is needed. Do you need atomic to manage the init?

I don't need since the initialization happen in the single thread environment. How about to use a mutex to protect the refcount?

arnopo:
> > @TanmayShah-xilinx: Thanks for the feedback I came to the same conclusion. The atomic management depends on OS, compiler but also processor. It just a nightmare to manage it in libmetal.... For Zephyr , a specific API is used (https://elixir.bootlin.com/zephyr/latest/source/include/zephyr/sys/atomic.h) The CI fails for a Cortex-M0.
> 
> I use atomic API because it is abstracted in libmetal.

Yes, the abstraction should do the job. Unfortunately if a standard API exists for atomic, there's no guarantee that a lib will be associated. This makes it complex to make generic.

> 
> > @xiaoxiang781216 : Regarding regression introduce I plan to remove this commit from the main branch. We have to rework the solution, if this fix is needed. Do you need atomic to manage the init?
> 
> I don't need since the initialization happen in the single thread environment. How about to use a mutex to protect the refcount?

yes, I had also mutex in mind, but I just wonder if it is not over protection. 
If you don't need it, I propose to make it simple in a first step by adding a simple counter.
If someone needs to protect from race condition, later,we could add it

xiaoxiang781216:
Ok, have you reverted my change? I can adjust the change on the current code base.

arnopo:
As your patch was the last I merged, i just remove it from the main branch without the revert mechanism

xiaoxiang781216:
Ok, I will create a new patcch with normal ++/-- today.

arnopo:
> Ok, I will create a new patcch with normal ++/-- today.

Thanks!

xiaoxiang781216:
@arnopo here is the new patch without atomic: https://github.com/OpenAMP/libmetal/pull/244

PR #237, opened 2023-05-30
Add Versal SOC family support

bentheredonethat:
Add support for Versal and Versal_Net SoCs APU cores A72 and A78

PR #242, opened 2023-07-10
lib: remove MB and GB definition in utilities.h

arnopo:
The introduction of MB and GB macros introduces warnings in Zephyr build.
Indeed they are already defined in zephyr/include/zephyr/sys/util.h. Remove them and put them back in
lib/system/generic/xlnx_common/zynqmp_aarch64/sys.c.

### DIFF
--- a/lib/system/generic/xlnx_common/zynqmp_aarch64/sys.c
+++ b/lib/system/generic/xlnx_common/zynqmp_aarch64/sys.c
@@ -27,6 +27,14 @@
 #include "xreg_cortexa53.h"
 #endif /* defined(versal) */
 
+#ifndef MB
+#define MB (1024 * 1024UL)
+#endif /* MB */
+
+#ifndef GB
+#define GB (MB * 1024UL)
+#endif /* GB */
+
 void sys_irq_restore_enable(unsigned int flags)
 {
 	Xil_ExceptionEnableMask(~flags);
--- a/lib/utilities.h
+++ b/lib/utilities.h
@@ -25,9 +25,6 @@ extern "C" {
  *  @{
  */
 
-#define MB (1024 * 1024UL)
-#define GB (1024 * 1024 * 1024UL)
-
 /** Marker for unused function arguments/variables. */
 #define metal_unused(x)	do { (x) = (x); } while (0)
 


arnopo:
@bentheredonethat:
This PR fix #237. Please, could you confirm that it works on your side

bentheredonethat:
@arnopo  confirmed

TanmayShah-xilinx (approved):
Looks good to me.

lib/system/generic/xlnx_common/zynqmp_aarch64/sys.c
```diff
@@ -27,6 +27,9 @@
 #include "xreg_cortexa53.h"
 #endif /* defined(versal) */
 
+#define MB (1024 * 1024UL)
```
TanmayShah-xilinx, line 30: Thanks @arnopo for this change.
Just to avoid any future warnings, we can do "#ifndef MB".
But, me or @bentheredonethat will do that change.

edmooring (approved):
Good to go.

bentheredonethat:
Hi @arnopo  I would actually push back on this patch

for ports where we have support for multiple processors (which we plan to upstream) then we would have to revert this patch because multiple processors would use the MB/GB symbols.

CC @tnmysh

lib/system/generic/xlnx_common/zynqmp_aarch64/sys.c
arnopo: done

arnopo:
> Hi @arnopo I would actually push back on this patch
> 
> for ports where we have support for multiple processors (which we plan to upstream) then we would have to revert this patch because multiple processors would use the MB/GB symbols.
> 
> CC @tnmysh

Hi @bentheredonethat 
Could I merge this one or do you plan to provide a better fix in a the short term?

tnmysh:
> > Hi @arnopo I would actually push back on this patch
> > for ports where we have support for multiple processors (which we plan to upstream) then we would have to revert this patch because multiple processors would use the MB/GB symbols.
> > CC @tnmysh
> 
> Hi @bentheredonethat Could I merge this one or do you plan to provide a better fix in a the short term?

@arnopo I will provide fix.

tnmysh:
@arnopo @bentheredonethat done: https://github.com/OpenAMP/libmetal/pull/249/

PR #243, opened 2023-07-10
CI: Fix checkpatch

arnopo:
The CI does not consider "check" reported by checkpatch. This commit fixes it by:

- rebasing check_compliance.py compliance.yml to integrate Zephyr updates,
- add detection of the "check" in the report.


arnopo:
Similar to https://github.com/OpenAMP/open-amp/pull/497
I merge it as it is CI test

PR #244, opened 2023-07-10
Add reference count to metal_common_state

xiaoxiang781216:
to allow to invoke metal_init and metal_finish multiple times

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -11,6 +11,9 @@ int metal_init(const struct metal_init_params *params)
 {
 	int error = 0;
 
+	if (_metal.common.ref_count++ != 0)
+		return 0;
+
 	memset(&_metal, 0, sizeof(_metal));
 
 	_metal.common.log_handler   = params->log_handler;
@@ -24,11 +27,15 @@ int metal_init(const struct metal_init_params *params)
 	if (error)
 		return error;
 
+	++_metal.common.ref_count;
 	return error;
 }
 
 void metal_finish(void)
 {
+	if (--_metal.common.ref_count != 0)
+		return;
+
 	metal_sys_finish();
 	memset(&_metal, 0, sizeof(_metal));
 }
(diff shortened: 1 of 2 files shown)


lib/init.c
```diff
@@ -11,24 +11,30 @@ int metal_init(const struct metal_init_params *params)
 {
 	int error = 0;
 
-	memset(&_metal, 0, sizeof(_metal));
+	if (_metal.common.ref_count++ == 0) {
+		memset(&_metal.common + 1, 0,
+		       sizeof(_metal) - sizeof(_metal.common));
```
arnopo, line 16: can be write on a single line ( 100 char/line)

arnopo (changes requested):

lib/init.c
```diff
@@ -11,24 +11,30 @@ int metal_init(const struct metal_init_params *params)
 {
 	int error = 0;
 
-	memset(&_metal, 0, sizeof(_metal));
+	if (_metal.common.ref_count++ == 0) {
+		memset(&_metal.common + 1, 0,
```
arnopo, line 15: `&_metal.common + 1 i`s probably not good as` ref_count `is an integer.
More than that I wonder if creating a static variable for refcount would not be more better than having this address shift

arnopo (changes requested):

lib/init.c
xiaoxiang781216: Done.

lib/init.c
xiaoxiang781216: but is it good to use global reference count for the local variables at least that it isn't a global variable in libmetal level?

xiaoxiang781216: Ok, I ajdust the change, please take a look.

arnopo (approved):
LGTM
thanks!

TanmayShah-xilinx (approved):
LGTM.

PR #245, opened 2023-07-10
lib: refactor alloc.h to not need forward declaration

glneo:
This should have no functional change, but makes this header consistent with how the other headers are organized.

### DIFF
--- a/lib/alloc.h
+++ b/lib/alloc.h
@@ -12,6 +12,8 @@
 #ifndef __METAL_ALLOC__H__
 #define __METAL_ALLOC__H__
 
+#include <metal/system/@PROJECT_SYSTEM@/alloc.h>
+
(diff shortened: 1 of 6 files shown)


arnopo:
Hi @glneo, 
Thanks to try to make libmetal API more consistent.
Could you share with us what led you to propose this PR. Did you encounter any particular problem?

glneo:
This is part of a larger set of refactoring that should culminate in the addition of support for TI devices.

The later changes make more sense and are easier to comprehend when the headers are more consistent. This change was sent separate as it should be trivially correct and can stand alone. Helps us test the waters with this project.

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #248, opened 2023-08-09
lib: system: freertos: Add metal/errno.h to match standalone

bentheredonethat:
As lib/system/generic/sys.h has metal/errno.h add this to the freertos equivalent.

This fixes some compilation issues when linking against certain FreeRTOS port BSPs.


arnopo:
@edmooring , @tnmysh 
could you review it please

tnmysh (approved):
LGTM

PR #249, opened 2023-08-16
lib: utilities: add MB and GB defines

tnmysh:
If MB and GB definition is not available then define new ones.

This also addresses: https://github.com/OpenAMP/libmetal/pull/242

### DIFF
--- a/lib/utilities.h
+++ b/lib/utilities.h
@@ -25,8 +25,13 @@ extern "C" {
  *  @{
  */
 
-#define MB (1024 * 1024UL)
-#define GB (1024 * 1024 * 1024UL)
+#ifndef MB
+#define MB (1024UL << 10UL)
+#endif
+
+#ifndef GB
+#define GB (MB << 10UL)
+#endif
 
 /** Marker for unused function arguments/variables. */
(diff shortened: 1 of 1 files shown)


lib/utilities.h
```diff
@@ -25,8 +25,13 @@ extern "C" {
  *  @{
  */
 
-#define MB (1024 * 1024UL)
-#define GB (1024 * 1024 * 1024UL)
+#ifndef MB
+#define MB (1024UL << 10UL)
```
tnmysh, line 29: @arnopo 

I have question here. Since MB and GB are so common terms, should we check values compile time of MB and GB if they are defined somewhere else ?

For example:

```
#ifndef MB
...
#else

#if (MB != (1024 * 1024))
#error "error: unexpected value of MB"
#endif

```

something like this? Since MB can be defined as 1000 * 1000. 

I am asking since want to know if it's overkill and we can simply assume that everywhere it will be defined as 1024 * 1024.

Thanks,
Tanmay

arnopo (approved):

arnopo:
@edmooring 
could you review it please  ?

PR #250, opened 2023-09-05
log: add ML_ERR, ML_INFO, ML_DBG macros

kernelchuk:
Convenience macros ML_ERR, ML_INFO, ML_DBG add source function name and the line number before the message.


kernelchuk:
I missed the copyright update.

arnopo:
I have difficulty seeing the objective of defining these macros in libmetal.
Is it only for the application, or do you also expect to use it in openamp lib?

arnopo:
I close it as addressed in #254

PR #252, opened 2023-09-11
Add xilinx a72 and a78

bentheredonethat:
Add support for xilinx platforms cortex a72 on Versal, cortex a78 on versal-net

both of these have support for standalone and baremetal


bentheredonethat:
Hi @arnopo  the failed check are due to Xilinx BSP. The failures are present in other platforms too.

Thanks,
Ben

arnopo:
Hi @bentheredonethat,

> Hi @arnopo the failed check is due to Xilinx BSP. The failures are present in other platforms too.
> 

I can see two issues in CI one is linked to the format of one of your patch, the second is linked to the Zephyr CI, that I just fixed. 

> We need the routines as part of enabling Xilinx BSP for these platforms.
> 
> Thanks, Ben

bentheredonethat:
@arnopo  ok i can update commist message due to warning
"1: UC5 Title exceeds max length (98>72): "lib: generic: zynqmp_a78: Add support for versal_net a78 platform with System Device Tree workflow"

was there anything else?

lib/system/freertos/xlnx_common/zynqmp_aarch64/sys.h
```diff
@@ -0,0 +1,46 @@
...
+extern "C" {
+#endif
+
+#ifdef METAL_INTERNAL
+
+#define XLNX_MAXIRQS XSCUGIC_MAX_NUM_INTR_INPUTS
+
+static inline void sys_irq_enable(unsigned int vector)
+{
+	XScuGic_EnableIntr(XPAR_SCUGIC_0_DIST_BASEADDR, vector);
```
tnmysh, line 32: Since this is under freertos directory is it possible to use FreeRTOS APIs to do this instead of using XILINX APIS. For example to enable irq, freertos must be providing some kind of API.

bentheredonethat: I looked in Xilinx BSP port of FReeRTOS and using vPortEnableInterrupt() might work. I will have to test.

lib/system/freertos/xlnx_common/zynqmp_aarch64/sys.c
```diff
@@ -0,0 +1,115 @@
...
+
+	return state;
+}
+
+void metal_machine_cache_flush(void *addr, unsigned int len)
+{
+	if (!addr || !len)
+		Xil_DCacheFlush();
+	else
+		Xil_DCacheFlushRange((intptr_t)addr, len);
```
tnmysh, line 55: Here too for cache flush. We can sync with developers who are working on Xilinx's freertos port to ask correct API that should be used to perform all operations in this file.

bentheredonethat: this part I am would push back for 2 reasons:
1. This API is already in libmetal for other ports. 
2. There is no alternative for these APIs in Xilinx FreeRTOS port.

tnmysh: That is what I am trying to say. We need to introduce cache support in FreeRTOS and then use those API here. So, metal_machine_cache_flush interface will remain same, but implementation should be freertos specific.

tnmysh: If FreeRTOS call isn't available, then that API doesn't belong in this directory. This directory is supposed to translate libmetal calls to freertos calls. Can't introduce baremtal calls to platforms here. Please move such APIs under generic/<platform>/directory.

tnmysh: For versal-net support, we need generic/zynqmp_r52 directory and have calls there.

edmooring:
@arnopo,
How did you handle this for the existing files like apps/system/generic/machine/zynqmp_r5/helper.c in your initial integration with checkpatch?

I could argue that checking for CamelCase in the platform-specific code is wrong in that this code may have to call down into BSP-specific code that may have a very different set of coding conventions.

arnopo:
> How did you handle this for the existing files like apps/system/generic/machine/zynqmp_r5/helper.c in your initial integration with checkpatch?
> 
> I could argue that checking for CamelCase in the platform-specific code is wrong in that this code may have to call down into BSP-specific code that may have a very different set of coding conventions.

I did not handle the exception, it could become difficult to manage. 
For me, failing the test but with a justification is acceptable.

lib/system/freertos/xlnx_common/zynqmp_aarch64/sys.c
tnmysh: Ok I checked FreeRTOS code, they don't provide any standard FreeRTOS specific framework for Cache Management instead, it is upto platforms how to implement it. So, in that case we have to use Xilinx BSP APIs here.

lib/system/freertos/xlnx_common/zynqmp_aarch64/sys.c
```diff
@@ -0,0 +1,115 @@
...
+inline void metal_weak metal_freertos_default_poll(void)
+{
+	metal_asm volatile("wfi");
+}
+
+/* TODO: document functionality of this function. */
+void *metal_machine_io_mem_map(void *va, metal_phys_addr_t pa,
+			       size_t size, unsigned int flags)
+{
+	unsigned long section_offset;
```
tnmysh, line 78: I prefer we don' have platform specific implementation in libmetal. Instead push this implementation in Xilinx's BSP and use only interfaces here from the BSP. That way, It's clean transformation from libmetal to BSP.

I see Xil_MemMap API is available to achieve mapping of memory for ZynqMP R5 platform so, Probably that API needs to be ported for other Xilinx Platforms' BSP.

bentheredonethat:
@tnmysh  to be clear - is the only gap to fix right now the metal-io-mem-map() part? the cache part there is already convention of other platforms.

if so i will work on updating Xilinx BSP and then update the PR with that accordingly.

tnmysh:
@arnopo we have refactored Xilinx platform specific files and added new platform support accordingly in   this PR. I think it's ready to be merged, we can have this one as part of this release.

Do you have any extra comments? I see some failures with checkpatch. camelCasing can't be fixed. Other than that this PR looks good to me.

tnmysh (approved):
LGTM.

lib/system/generic/xlnx_common/sys.c
```diff
@@ -1,26 +1,37 @@
...
- * Copyright (C) 2022, Advanced Micro Devices, Inc.
+ * Copyright (c) 2022, Xilinx Inc. and Contributors. All rights reserved.
+ * Copyright (c) 2022-2023 Advanced Micro Devices, Inc. All Rights Reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
 /*
- * @file	generic/zynqmp_r5/sys.c
+ * @file	generic/xlnx_common/zynqmp_aarch64/sys.c
```
arnopo, line 9: generic/xlnx_common/sys.c

lib/system/generic/xlnx_common/zynqmp_aarch64/sys.c
```diff
@@ -19,13 +19,17 @@
 #include "xil_mmu.h"
 #include "xscugic.h"
 
+/* System Device Tree (SDT) flow does not have the files generated. */
```
arnopo, line 23: this file seems not build the CmakeLists.txt has been deleted.

lib/system/freertos/CMakeLists.txt
```diff
@@ -21,4 +21,12 @@ if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_MACHINE})
   add_subdirectory(${PROJECT_MACHINE})
 endif (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_MACHINE})
 
+if ("${PROJECT_MACHINE}" STREQUAL "zynqmp_a53" OR
+    "${PROJECT_MACHINE}" STREQUAL "zynqmp_a72" OR
+    "${PROJECT_MACHINE}" STREQUAL "zynqmp_a78" OR
+    "${PROJECT_MACHINE}" STREQUAL "zynqmp_r5" OR
+    "${PROJECT_MACHINE}" STREQUAL "zynq7")
+add_subdirectory(xlnx_common)
```
arnopo, line 29: The folder name doesn't seem maintainable in long term.
what about Xilinx or AMD, then in case of new AMD platforms some subfolders could be added

lib/system/freertos/sys.h
```diff
@@ -19,8 +19,6 @@
 
 #include <metal/errno.h>
 
-#include "./@PROJECT_MACHINE@/sys.h"
```
arnopo, line 22: Could you explain why you remove this?
This probably break the build with the template or another machine

arnopo (changes requested):

lib/system/freertos/sys.h
tnmysh: Hi @arnopo, your concern is valid. but that line forces us to duplicate sys.h interfaces for each platforms. Instead, can we move sys.h interfaces only into freertos/sys.h and then each platform's sys.c should implement those interfaces. This will reduce duplicate sys.h files for each platforms. For example, Only for Xilinx Platforms, we have about 6 duplicate sys.h files that contains same interfaces.

lib/system/freertos/CMakeLists.txt
tnmysh: @arnopo it's not decided yet. For now I suggest we keep vendor name as xlnx. Also, Platform BSP API used for xlnx platforms and AMD platforms will be different, so we will need separate directory for them anyway.

lib/system/generic/sys.h
```diff
@@ -23,7 +23,9 @@
 #include <stdarg.h>
 #include <string.h>
 
+#ifdef __MICROBLAZE__
 #include "./@PROJECT_MACHINE@/sys.h"
+#endif
```
arnopo, line 28: I missed this one too. this breaks generic platform implementation

lib/system/freertos/CMakeLists.txt
tnmysh: Actually I prefer we keep one directory only for Xilinx's platforms and second directory for AMD's platforms as they both will have separate BSPs and so API used to implement interfaces will be different too.

If we want to avoid xlnx vendor name, We can go ahead with "zynqmp_common" name, as it is "zynqmp" family platforms.

@arnopo are you ok with that?

lib/system/freertos/sys.h
arnopo: I understand the redundancy, This is linked to the way the libmetal is designed. 
What we have to keep in mind here is that this will generate regression on projects that use the Libmetal  generic or Freertos environments. 
So Instead of removing this for all platforms you should make an exception for the Xilinx platforms

arnopo (changes requested):

bentheredonethat:
Hi @arnopo i just pushed another version. i testd zynq, zynqmp-a53 and zynqmp-r5 for freertos and generic. all of these passed.

lib/system/generic/sys.h
bentheredonethat: just asking, why does this break the implementation? 

There is ifdef check for microblaze BSP here

arnopo: For instance, you break the "template" machine, and all machines that are not upstreamed.
Then anyone who wants to upstream their machine would have to add their own `#ifdef` statement

tnmysh: Ok, we can have check only for Xilinx's platform and keep the Machine specific header.

bentheredonethat: updated this @arnopo  .. please see latest patches

lib/system/generic/xlnx/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+  add_subdirectory(microblaze_generic)
+else()
+  collect (PROJECT_LIB_SOURCES sys.c)
+endif()
+
+collect (PROJECT_LIB_HEADERS sys.h)
+
+collect (PROJECT_LIB_SOURCES irq.c)
+
+# vim: expandtab:ts=2:sw=2:smartindent
```
arnopo, line 11: remove this line

arnopo (approved):
Just a minor clean-up to address, else LGTM

lib/system/generic/xlnx/CMakeLists.txt
bentheredonethat: which line? the line 9 is required for xlnx platforms
@arnopo

bentheredonethat:
Hi @arnopo  on my end there is a lot of testing to cover. I would ask that this not be included in the upcoming release.

Thanks
CC @tnmysh

tnmysh:
@arnopo I removed my ACK till Ben confirms this is working.
We can exclude this commit from this release and merge it later.

lib/system/generic/xlnx/CMakeLists.txt
arnopo: line 11:` # vim: expandtab:ts=2:sw=2:smartindent`

bentheredonethat:
@arnopo i did testing for builds of apps and demos that use libmetal for xilinx paltforms. im ok t to proceed now
CC @tnmysh

tnmysh (approved):
This patch removes almost 1000 lines (~400 line of code) from libmetal library and we can achieve same platform support. 

Thanks @bentheredonethat for efforts. When your side testing is done please let us know.
This will be good addition to release.

bentheredonethat:
im ok with testing on my side

tnmysh:
> @bentheredonethat Please, could you rebase it on top of main branch that fixes CI test
> 
> @tnmysh : then ok to merge?

Yes I am ok to merge.

bentheredonethat:
@arnopo  rebased off latest

PR #254, opened 2023-09-29
log: add function name, line number prefix

kernelchuk:
See also https://github.com/OpenAMP/libmetal/pull/250
Add "function-name:line-number" prefix and a new-line suffix for all messages. Use do-while(0) to define the metal_log macro. Add convenience macros ML_ERR, ML_WRN, ML_INF, ML_DBG to avoid using excessively long and redundant "metal_log(METAL_LOG_*".

I had to change s/irq/IRQ/ to fix the checkpatch warning: 
`EMBEDDED_FUNCTION_NAME: Prefer using '"%s...", __func__' to using 'irq', this function's name, in a string`

### DIFF
--- a/lib/log.h
+++ b/lib/log.h
@@ -72,15 +72,38 @@ extern void metal_default_log_handler(enum metal_log_level level,
 				      const char *format, ...);
 
 /**
- * Emit a log message if the log level permits.
+ * @internal
+ *
+ * @brief	used by the metal_log() macro to update the format string
+ *
+ * If ML_FUNC_LINE is defined this macro generates a unified format
+ * string for metal_log() and its convenience metal_*() macros, i.e. it
+ * adds function-name:line-number prefix to all log messages.
+ *
+ * @param[in]	fmt	format string passed from the metal_log() macro
+ */
+#if defined(ML_FUNC_LINE)
+#define metal_fmt(fmt) "%s:%u " fmt, __func__, __LINE__
+#else	/* ML_FUNC_LINE */
+#define metal_fmt(fmt) fmt
+#endif	/* ML_FUNC_LINE */
+
+/**
+ * @brief	Emit a log message if the log level permits.
  *
  * @param	level	Log level.
- * @param	...	Format string and arguments.
+ * @param	fmt	Format string.
+ * @param	args... Variable number of arguments.
  */
-#define metal_log(level, ...)						       \
-	((level <= _metal.common.log_level && _metal.common.log_handler) \
-	       ? (void)_metal.common.log_handler(level, __VA_ARGS__)	       \
-	       : (void)0)
+#define metal_log(level, fmt, args...) ({				   \
+	if (_metal.common.log_handler && level <= _metal.common.log_level) \
+		_metal.common.log_handler(level, metal_fmt(fmt), ##args);  \
+})
+
+#define metal_err(fmt, args...) metal_log(METAL_LOG_ERROR, fmt, ##args)
+#define metal_warn(fmt, args...) metal_log(METAL_LOG_WARNING, fmt, ##args)
+#define metal_info(fmt, args...) metal_log(METAL_LOG_INFO, fmt, ##args)
+#define metal_dbg(fmt, args...) metal_log(METAL_LOG_DEBUG, fmt, ##args)
 
 /** @} */
 
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -69,6 +69,7 @@ if (WITH_ZEPHYR)
 endif (WITH_ZEPHYR)
 
 option (WITH_DEFAULT_LOGGER "Build with default logger" ON)
+option (WITH_FUNC_LINE_LOG "Log with function name, line number prefix" OFF)
 
 option (WITH_DOC "Build with documentation" ON)
 
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -65,6 +65,10 @@ if (WITH_DEFAULT_LOGGER)
   add_definitions (-DDEFAULT_LOGGER_ON)
 endif (WITH_DEFAULT_LOGGER)
 
+if (WITH_FUNC_LINE_LOG)
+  add_definitions (-DML_FUNC_LINE)
+endif (WITH_FUNC_LINE_LOG)
+
 get_property (_ec_flgs GLOBAL PROPERTY "PROJECT_EC_FLAGS")
 
 if (WITH_ZEPHYR)


arnopo:
@kernelchuk do  you have some measurement of the increase of the footprint adding this?

Any reason to not merge this PR with https://github.com/OpenAMP/libmetal/pull/250?

kernelchuk:
> Any reason to not merge this PR with #250?

None. I just pushed a new branch as a replacement for #250. Should I go back to #250?

> @kernelchuk do you have some measurement of the increase of the footprint adding this?

I estimate the upper bound to be less than 1000 bytes:
``` 
total len M     function_name 
===== === = =======================; M = number of log messages
   23  23 1 metal_bus_register
   48  25 1 metal_bus_unregister
   72  24 1 metal_get_timestamp
   98  26 1 metal_init_page_sizes
  125  27 1 metal_linux_irq_notify
  146  21 1 metal_shmem_open
  173  27 1 metal_softirq_allocate
  199  26 1 metal_uio_dev_dma_map
  261  31 2 metal_linux_irq_set_enable
  319  29 2 metal_linux_probe_driver
  371  26 2 metal_uio_dev_irq_ack
  411  20 2 metal_virt2phys
  459  24 2 metal_xlnx_irq_init
  519  30 2 metal_xlnx_irq_set_enable
  591  24 3 metal_add_page_size
  666  25 3 metal_linux_irq_init
  753  29 3 metal_linux_irq_shutdown
  804  17 3 metal_mktemp
  876  24 3 metal_shmem_try_map
  952  19 4 metal_sys_init
 1097  29 5 metal_linux_irq_handling
 1235  23 6 metal_uio_dev_bind
 1419  23 8 metal_uio_dev_open
```
Where 'total' is a running total of bytes; 'len' is the length of "function-name:line-number"; M is the number of log messages in the function. I'm assuming that all line numbers are 3 digits long, so counting as 5 for ":nnn ". We get to under 1000 bytes is we adjust for `__func__` strings in the following functions:
```
total len N     function_name
===== === = =======================; N = number of __func__ in the replaced messages
   24  24 1 metal_linux_irq_notify
   50  26 1 metal_linux_irq_shutdown
   73  23 1 metal_uio_dev_dma_map
   93  20 1 metal_uio_dev_open
  149  28 2 metal_linux_irq_set_enable
  201  26 2 metal_linux_probe_driver
  247  23 2 metal_uio_dev_irq_ack
  289  21 2 metal_xlnx_irq_init
  343  27 2 metal_xlnx_irq_set_enable
  473  26 5 metal_linux_irq_handling
```

lib/log.h
```diff
@@ -77,10 +77,17 @@ extern void metal_default_log_handler(enum metal_log_level level,
...
-	       ? (void)_metal.common.log_handler(level, __VA_ARGS__)	       \
-	       : (void)0)
+#define metal_log(level, fmt, args ...)					\
+do {									\
+	if (_metal.common.log_handler && level <= _metal.common.log_level) \
+		_metal.common.log_handler(level, "%s:%u " fmt "\n",	\
+					  __func__, __LINE__, ##args);	\
+} while (0)
+
+#define ML_ERR(fmt, args ...) metal_log(METAL_LOG_ERROR, fmt, ##args)
```
tnmysh, line 87: Can we have macros like ml_err, ml_info, ml_debug, ml_warn.
These names are matching with kernel pr_* macros.
It will be easy for users to stay align with names that are matching what kernel has.

lib/log.h
```diff
@@ -77,10 +77,17 @@ extern void metal_default_log_handler(enum metal_log_level level,
...
-#define metal_log(level, ...)						       \
-	((level <= _metal.common.log_level && _metal.common.log_handler) \
-	       ? (void)_metal.common.log_handler(level, __VA_ARGS__)	       \
-	       : (void)0)
+#define metal_log(level, fmt, args ...)					\
+do {									\
+	if (_metal.common.log_handler && level <= _metal.common.log_level) \
+		_metal.common.log_handler(level, "%s:%u " fmt "\n",	\
+					  __func__, __LINE__, ##args);	\
+} while (0)
```
tnmysh, line 85: So, I checked size of libmetal before and after this patch with this command: "size libmetal.a"

I see before .text section: 25187 i.e. ~25.1K
After .text section: 23958 i.e. ~23.9K. 
~1.2k difference.

Macros are certainly adding values. But, __func__, __LINE__ and "\n" forcing in all the logs may not be appreciated by every user.
I prefer we leave that choice to user.

arnopo: Agree with @tnmysh 
 One of the pillars of the library is to be able to run with a minimal footprint. If the addition of functions name and lines should impact the nominal footprint this should be something that is disable by default.

edmooring (commented):
This should be conditional. Many deployments of OpenAMP will not have any log output visible to the user, so adding extra size is not useful for them.

kernelchuk:
> This should be conditional. Many deployments of OpenAMP will not have any log output visible to the user, so adding extra size is not useful for them.

Yes, I'm working on a version to make it conditional.

kernelchuk:
Hi @arnopo,
I got the `openamp-docs-libmetal-prs` build error:
```
The configuration file required to build documentation is missing from your project.
Add a configuration file to your project to make it build successfully. Read more at
https://docs.readthedocs.io/en/stable/config-file/v2.html
```
I have scanned the readthedocs.io link, but still not sure how to avoid it. Any hints?

arnopo:
> Hi @arnopo, I got the `openamp-docs-libmetal-prs` build error:
> 
> ```
> The configuration file required to build documentation is missing from your project.
> Add a configuration file to your project to make it build successfully. Read more at
> https://docs.readthedocs.io/en/stable/config-file/v2.html
> ```
> 
> I have scanned the readthedocs.io link, but still not sure how to avoid it. Any hints?

@wmamills : I guess we can ignore the doc generation issue for now, right?

wmamills:
@arnopo, yes ignore it until that enabling PR is accepted.  The hook is installed already but the needed file is not yet in main.

lib/softirq.c
```diff
@@ -72,8 +72,7 @@ int metal_softirq_allocate(int num)
 	int irq_base;
 
 	if ((metal_softirq_avail + num) >= metal_softirq_num) {
-		metal_log(METAL_LOG_ERROR, "No %d available soft irqs.\r\n",
-			  num);
+		ml_err("No %d available soft irqs.\r\n", num);
```
arnopo, line 75: Do you need to replace` metal_log(METAL_LOG_ERROR, ...) `by `ml_err `?

kernelchuk: No. Please, see the new version.

lib/log.h
```diff
@@ -71,16 +71,36 @@ extern enum metal_log_level metal_get_log_level(void);
...
 				      const char *format, ...);
 
+/**
+ * @brief	used by the metal_log() macro to update the format string
+ * @fmt	format string passed from the metal_log() macro
+ *
+ * If ML_FUNC_LINE is defined this macro generates a unified format
+ * string for metal_log() and its convenience ml_*() macros, i.e. it
+ * adds function-name:line-number prefix to all log messages.
+ */
```
arnopo, line 81:
```c
/**
 * @internal 
 *
 * @brief	used by the metal_log() macro to update the format string
 *
 * If ML_FUNC_LINE is defined this macro generates a unified format
 * string for metal_log() and its convenience ml_*() macros, i.e. it
 * adds function-name:line-number prefix to all log messages.
 *
 * @param[in] fmt	format string passed from the metal_log() macro
 */
```

lib/log.h
```diff
@@ -71,16 +71,36 @@ extern enum metal_log_level metal_get_log_level(void);
...
+#else	/* ML_FUNC_LINE */
+#define ml_fmt(fmt) fmt
+#endif	/* ML_FUNC_LINE */
+
 /**
  * Emit a log message if the log level permits.
  *
  * @param	level	Log level.
  * @param	...	Format string and arguments.
  */
```
arnopo, line 93:
```c
/**
 * @ brief Emit a log message if the log level permits.
 *
 * @param	level	Log level.
 * @param	fmt	format string.
 * @param	args 	format string arguments.
 * @param	...	format string and arguments.
 */
```

lib/log.h
```diff
@@ -71,16 +71,36 @@ extern enum metal_log_level metal_get_log_level(void);
...
-#define metal_log(level, ...)						       \
-	((level <= _metal.common.log_level && _metal.common.log_handler) \
-	       ? (void)_metal.common.log_handler(level, __VA_ARGS__)	       \
-	       : (void)0)
+#define metal_log(level, fmt, args ...)					\
```
arnopo, line 94: ',' between `args `and `...`?

lib/log.h
```diff
@@ -71,16 +71,36 @@ extern enum metal_log_level metal_get_log_level(void);
...
+do {									\
+	if (_metal.common.log_handler && level <= _metal.common.log_level) \
+		_metal.common.log_handler(level, ml_fmt(fmt), ##args);	\
+} while (0)
```
arnopo, line 98: avoid while loop.

#define metal_log(level, fmt, args ...)  ( {					\
	if (_metal.common.log_handler && level <= _metal.common.log_level) 	\
		_metal.common.log_handler(level, ml_fmt(fmt), ##args);		\
})

lib/log.h
```diff
@@ -71,16 +71,36 @@ extern enum metal_log_level metal_get_log_level(void);
...
+
+#define ml_err(fmt, args ...) metal_log(METAL_LOG_ERROR, fmt, ##args)
+#define ml_warn(fmt, args ...) metal_log(METAL_LOG_WARNING, fmt, ##args)
+#define ml_info(fmt, args ...) metal_log(METAL_LOG_INFO, fmt, ##args)
+#define ml_dbg(fmt, args ...) metal_log(METAL_LOG_DEBUG, fmt, ##args)
```
arnopo, line 103: These macros seem useless

arnopo (changes requested):
Much better§
Few new remarks

lib/log.h
kernelchuk: We have discussed these in our RP call and agreed to add "convenience macros ml_err, ml_warn, ml_info, ml_dbg to avoid using excessively long and redundant "metal_log(METAL_LOG_*".

tnmysh: @arnopo I see values in these macros. Developers can use short macros with clean interface for logging. Instead of metal_log(METAL_LOG_* it is cleaner to use ml_* macros which aligns kernel's pr_* macros. So, we can keep macros and new logs can be added with these macros.

Having said that, we can replace ml_dbg with ml_debug.

arnopo: That's valid. 
If we keep them, we have to find a naming started by "metal_" to be consistent.
what about metal_err, metal_warb, metal_info, metal_debug (or metal_dbg)?

lib/log.h
kernelchuk: Thank you. I'll update.

lib/log.h
kernelchuk: Thank you. I'll update.

lib/log.h
kernelchuk: Adding ',' would result in compilation errors, e.g.:
```
error: macro "metal_log" requires 4 arguments, but only 2 given
   | metal_log(METAL_LOG_DEBUG, "failed to allocate memory\n");
   |                                                         ^
```
Please, see "Macros with a Variable Number of Arguments" section in the [gcc manual](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst).

lib/log.h
kernelchuk: I'm following the kernel [coding-style](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst) section-12 guidance:
_Macros with multiple statements should be enclosed in a do - while block_
I can change it.

lib/log.h
kernelchuk: If changing form ml_err to metal_err, etc. is a condition of acceptance, I'd like to use metal_dbg. We should not impose higher typing, space cost if we want to encourage more messages for error reporting, debugging, etc. Here is the length (cost) of each version:
```
26 metal_log(METAL_LOG_DEBUG,
12 metal_debug(
10 metal_dbg(
 7 ml_dbg(
```
I'd vote for the shorter version. In this context "ml" stands for metal_log or metal and I can comment in the log.h.

lib/log.h
arnopo: Ok, I see that you suppress space between `args` and  `...` in your last version

lib/log.h
arnopo: Yes please, we not necessary follow the linux coding style. I would prefer to not continue to use `while(0)`, if an alternative exists.

kernelchuk: Thank you, I'll change to a compound statement `({ .. })` as you suggested.

lib/log.h
kernelchuk: Thank you.

lib/log.h
tnmysh: I am okay with metal_dbg as well. Please do the change accordingly.

kernelchuk: Thank you. Done.

arnopo (approved):

tnmysh (approved):

tnmysh:
@arnopo If all looks good, can we include the change in coming release as well?

tnmysh:
@kernelchuk Please close #250 as it's duplicate of this.

PR #256, opened 2023-10-07
Docs: enable readthedocs rendering (for PRs)

wmamills:
The Readthedocs pull-request integration is nice to have for PR review. Run the whole OpenAMP-docs build but use the PR's commit

This also requires a ReadTheDocs project to install its integration for this repo. The project should enable "Build pull request for this project". It is currently in the Advanced Settings panel.

The project will also run for the main branch but that should be ignored and only the main project URL should be shown outside of PRs.

To this end the sub-module project names will use "-prs" in the name.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,35 @@
+version: 2
+formats: all
+build:
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.10"
+  apt_packages:
+    - cmake
+    - libhugetlbfs-dev
+    - libsysfs-dev
+  # here it would be nice to tell them that the conf.py will be in openamp-docs/conf.py
+  # sphinx:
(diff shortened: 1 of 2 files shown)


wmamills:
This is now working on my libmetal fork.
I sent myself a PR to test it which you can see here:
https://github.com/wmamills/libmetal/pull/2

The read the docs project is already created and the webhook is already installed for this repo.
However, the RTDS check will fail until this PR is accepted.

After we are satisfied with libmetal, it is easy to setup for the other submodules of openamp-docs.

wmamills:
@arnopo any objection to accepting this now so we can start using it and stop getting false CI failures?

arnopo:
> @arnopo any objection to accepting this now so we can start using it and stop getting false CI failures?

Indeed, it is better to integrate it quickly.

PR #258, opened 2023-10-10
README: Point people to the offical docs not the wiki

wmamills:
The content from the wiki has been moved to the Sphinx based OpenAMP-docs. (Meeting notes will continue to be in the wiki only.)

The wiki data is probably out of date and will get more out of date as we go.

So point people to the official docs.
(Separately we plan to delete the non-meeting note wiki pages.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ and request memory across the following operating environments:
   * Bare-metal environments
 
 For more details on the framework please refer to the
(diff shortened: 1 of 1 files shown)


wmamills:
BTW: The doc PR review worked for this PR.

tammyleino (approved):

arnopo (approved):

PR #259, opened 2023-10-10
Docs: fix for the readthedocs PR rendering

wmamills:
The original procedure worked only when the PR was coming from the same repo that was the target of the PR.  This is not normally the case.

The original check out has the commit we need.  Just use it as the remote, fetch and checkout the commit hash we need.


wmamills:
OK I have done as much testing as I can.  I used two different forks of libmetal and sent a PR from one to the other.
See https://github.com/wmamills/libmetal/pull/3

What I found was we need the new .readthedocs.yaml (the one from this fix)
AND the fixed .readthedocs.yaml needs to be in the PR (I was expecting it to use main for that).

So what this means is THIS PR will pass (and it did above).
Any other PR will FAIL until the PR is accepted and the other PR rebased ontop.

Sorry the history is messy but hopefully it is OK now.

wmamills:
Once this is accepted I will rebase my fix the wiki PR and resubmit it and we can see this in action for real.

PR #261, opened 2023-10-12
github action: add stales action

arnopo:
Add action to add stale label on issues and PR when no activity during 45 days.


arnopo:
reviewed in [openamp](https://github.com/OpenAMP/open-amp/pull/512)

PR #262, opened 2023-10-12
cmake: fix native build on arm

bene42:
uname -m on 32bit Arm platforms (little endian) report a machine string like armv5tejl, armv7l or armv8l.

A native build on these machines fail:

CMake Error at lib/processor/CMakeLists.txt:1 (add_subdirectory):
  add_subdirectory given source "armv8l" which is not an existing directory.

Set PROJECT_PROCESSOR to arm on these machines.


arnopo:
Hello @bene42 
Do you have some examples, how other project manage this?
Look like Zephyr use ARCH `set(CMAKE_SYSTEM_PROCESSOR ${ARCH})`

I wonder if we should not use it also instead of having an exhaustive list.
could you confirm as a first step the value of ${ARCH} in your build environment?

bene42:
> Hello @bene42 Do you have some examples, how other project manage this? Look like Zephyr use ARCH `set(CMAKE_SYSTEM_PROCESSOR ${ARCH})`
> 
> I wonder if we should not use it also instead of having an exhaustive list. could you confirm as a first step the value of ${ARCH} in your build environment?

I look into that, but it take some time. I am more familiar with the autotools universe rather then cmake.
From that universe I can report that there is a rather huge switch-case logic involved. Maybe cmake hides it better... :)

PR #264, opened 2023-10-13
Remove vim editor configuration in files

arnopo:
Remove all remaining instances of "# vim: ..." in files


tnmysh (commented):
I am okay with the change.

However, do we know why they were introduced in the first place? 
May be we can give reasons in commit message and then remove it.

arnopo:
> I am okay with the change.
> 
> However, do we know why they were introduced in the first place? May be we can give reasons in commit message and then remove it.

This is part of the initial commit.  I suppose that the author(s) of the first verion of libmetal was using vim or gvim, which automatically add this line...

tammyleino (approved):

tnmysh (approved):
LGTM.

PR #266, opened 2023-10-15
Minor improvements in entrypoint CI script

aescolar:
A few minor changes to the CI script to reduce considerably the log size and do a minor speed up.
There are 3 main changes:
* wget prints less as it progresses (one dot per megabyte instead of KB) (~10K lines saved)
* The zephyr SDK extraction, instead of printing one line per file (~50K lines), it prints a progress bar updated every 3 seconds.
* west does a narrow update instead of a full update (only fetching the needed branch for each repo)

### DIFF
--- a/.github/actions/build_ci/entrypoint.sh
+++ b/.github/actions/build_ci/entrypoint.sh
@@ -44,7 +44,7 @@ build_generic(){
 build_freertos(){
 	echo  " Build for freertos OS "
       	apt-get install -y gcc-arm-none-eabi unzip &&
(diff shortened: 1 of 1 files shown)


arnopo:
Thanks for he improvement:  2637 line in log instead of more than 78000.
Could you please re-base it on main , that  integrates the fix for the Zephyr CI?

aescolar:
> Thanks for he improvement: 2637 line in log instead of more than 78000. Could you please re-base it on main , that integrates the fix for the Zephyr CI?

You are very welcome. It has been rebased and CI is green now :)

PR #265, opened 2023-10-13
test: Zephyr: Replace k_mem_block usage

arnopo:
The k_mem_block structure is obsolete for some years replace it by a "void *" pointer. as done in Zephyr.

This PR fixes following CI error for Zephyr tests:
```
/github/workspace/test/system/zephyr/main.c:19:20: error: array type has incomplete element type 'struct k_mem_block'
78014
   19 | struct k_mem_block block[BLK_NUM_MAX];
78015
      |                    ^~~~~
78016
/github/workspace/test/system/zephyr/main.c: In function 'metal_zephyr_allocate_memory':
78017
/github/workspace/test/system/zephyr/main.c:36:25: error: invalid use of undefined type 'struct k_mem_block'
78018
   36 |                 if (!blk->data) {
78019
      |                         ^~
78020
/github/workspace/test/system/zephyr/main.c:37:28: error: invalid use of undefined type 'struct k_mem_block'
78021
   37 |                         blk->data = k_heap_alloc(&kmpool, size, K_NO_WAIT);
78022
      |                            ^~
78023
/github/workspace/test/system/zephyr/main.c:38:33: error: invalid use of undefined type 'struct k_mem_block'
78024
   38 |                         if (!blk->data)
78025
      |                                 ^~
78026
/github/workspace/test/system/zephyr/main.c:40:35: error: invalid use of undefined type 'struct k_mem_block'
78027
   40 |                         return blk->data;
78028
      |                                   ^~
78029
/github/workspace/test/system/zephyr/main.c: In function 'metal_zephyr_free_memory':
78030
/github/workspace/test/system/zephyr/main.c:55:24: error: invalid use of undefined type 'struct k_mem_block'
78031
   55 |                 if (blk->data == ptr) {
78032
      |                        ^~
78033
/github/workspace/test/system/zephyr/main.c:57:28: error: invalid use of undefined type 'struct k_mem_block'
78034
   57 |                         blk->data = NULL;
78035
      |                            ^~
78036
make[2]: *** [CMakeFiles/app.dir/build.make:76: CMakeFiles/app.dir/test/system/zephyr/main.c.obj] Error 1
78037
make[1]: *** [CMakeFiles/Makefile2:2779: CMakeFiles/app.dir/all] Error 2
```

PR #267, opened 2023-10-15
Fix Zephyr testcase with latest Zephyr

aescolar:
In Zephyr PR #63351 , commit:
a35e40f410750f3205f0d87489aa0dfa4006e1d3
Support for k_mem_block was removed which was used by this Zephyr test.

Update the tets to just use a void* (which is what should be used instead).

----

Note this is the cause for the current CI failures.


aescolar:
@arnopo I can see you actually queued a very fix in https://github.com/OpenAMP/libmetal/pull/265
So feel free to close this one if you prefer your fix :)

arnopo (approved):
I'm merging your fix it is more elegant

PR #270, opened 2023-10-24
lib: compiler support of arm compiler v6

arnopo:
The predefined macro __CC_ARM available on 5.x is migrated to the macro __arm__ for the version 6.x. Add defines to support both versions.

See https://developer.arm.com/documentation/ka005069/1-0/?lang=en for details


tnmysh (approved):
LGTM.

ldaj00 (commented):
LGTM

PR #276, opened 2023-11-13
zephyr: use k_cpu_idle instead of arch_cpu_idle

dcpleung:
arch_cpu_idle() is an internal API that should not be used outside of the main Zephyr. Use k_cpu_idle() instead, which is the same thing (at least for now).


arnopo (approved):
LGTM

edmooring (approved):
Looks good to go.

PR #279, opened 2023-12-04
libmetal/nuttx/io.c: width matched access when read/write size = 1,2,4,8

CV-Bowen:
libmetal/nuttx/io.c: width matched access when read/write size = 1,2,4,8

Follow the virtio spec v1.2
> 4.2.2.2 Driver Requirements: MMIO Device Register Layout:
The driver MUST only use 32 bit wide and aligned reads and writes to access the control registers described in table 4.1. For the device-specific configuration space, the driver MUST use 8 bit wide accesses for 8 bit wide fields, 16 bit wide and aligned accesses for 16 bit wide fields and 32 bit wide and aligned accesses for 32 and 64 bit wide fields.

So change the nuttx implemented io read/write operation.

### DIFF
--- a/lib/system/nuttx/io.c
+++ b/lib/system/nuttx/io.c
@@ -37,7 +37,17 @@ static int metal_io_block_read_(struct metal_io_region *io,
 	void *va = metal_io_virt(io, offset);
 
 	metal_cache_invalidate(va, len);
-	memcpy(dst, va, len);
+	if (len == 1)
+		*(uint8_t *)dst = *(uint8_t *)va;
+	else if (len == 2)
+		*(uint16_t *)dst = *(uint16_t *)va;
+	else if (len == 4)
+		*(uint32_t *)dst = *(uint32_t *)va;
+	else if (len == 8) {
+		*(uint32_t *)dst = *(uint32_t *)va;
+		*((uint32_t *)dst + 1) = *((uint32_t *)va + 1);
+	} else
+		memcpy(dst, va, len);
 
 	return len;
 }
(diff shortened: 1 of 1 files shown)


xiaoxiang781216 (approved):

arnopo:
If constraint is generic what about using macros metal_io_write/read<x>   ( x = 8,16,32 or 64) in open-amp?

CV-Bowen:
@arnopo Yes, we used libmetal implemented `metal_io_write/read()` before, but we met a bug when using atomic operation to operate the I/O space in risc-v in QMEU 8.1.2. CPU will hang at the atomic instruction.
See: https://github.com/apache/nuttx/pull/11201

CV-Bowen:
> > @arnopo Yes, we used libmetal implemented `metal_io_write/read()` before, but we met a bug when using atomic operation to operate the I/O space in risc-v in QMEU 8.1.2. CPU will hang at the atomic instruction. See: [apache/nuttx#11201](https://github.com/apache/nuttx/pull/11201)
> 
> Thanks for the link!
> 
> Something seems missing in you patch. You should check the alignement of the address to respect the virtio specification

@arnopo How about let's leave virtio devices and drivers to ensure that the address is aligned?

CV-Bowen:
@arnopo And could you check why the CI failed, seems not related to this PR?

lib/system/nuttx/io.c
```diff
@@ -37,7 +37,17 @@ static int metal_io_block_read_(struct metal_io_region *io,
...
-	memcpy(dst, va, len);
+	if (len == 1)
+		*(uint8_t *)dst = *(uint8_t *)va;
+	else if (len == 2)
+		*(uint16_t *)dst = *(uint16_t *)va;
+	else if (len == 4)
+		*(uint32_t *)dst = *(uint32_t *)va;
+	else if (len == 8) {
+		*(uint32_t *)dst = *(uint32_t *)va;
+		*((uint32_t *)dst + 1) = *((uint32_t *)va + 1);
```
arnopo, line 48: what about using uint64_t ?

CV-Bowen: The spec says:
The driver MUST use 8 bit wide accesses for 8 bit wide fields, 16 bit wide and aligned accesses for 16 bit wide fields and **32 bit wide and aligned accesses for 32 and 64 bit wide fields.**

arnopo (approved):

tnmysh (approved):

PR #280, opened 2023-12-14
Zephyr integration: Use K_HEAP_MEM_POOL_SIZE

jhedberg:
This define should be used instead of the Kconfig option to determine if the k_malloc family of functions are available.


jhedberg:
This is needed for https://github.com/zephyrproject-rtos/zephyr/pull/65908

arnopo (approved):
LGTM

carlocaione (approved):

PR #281, opened 2023-12-14
Latest Xilinx-AMD System Device Tree flow update

bentheredonethat:
This patch enables library and demo builds using latest Xilinx-AMD System Device (SDT) Flow update.

### DIFF
--- a/lib/system/generic/xlnx/CMakeLists.txt
+++ b/lib/system/generic/xlnx/CMakeLists.txt
@@ -8,4 +8,8 @@ collect (PROJECT_LIB_HEADERS sys.h)
 
 collect (PROJECT_LIB_SOURCES irq.c)
 
-# vim: expandtab:ts=2:sw=2:smartindent
+include(CheckSymbolExists)
+check_symbol_exists(SDT "bspconfig.h" HAS_SYSTEM_DT)
+if (HAS_SYSTEM_DT)
+  collect (PROJECT_LIB_HEADERS sys_devicetree.h)
+endif()
--- a/lib/system/generic/xlnx/sys.h
+++ b/lib/system/generic/xlnx/sys.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2015, Xilinx Inc. and Contributors. All rights reserved.
+ * Copyright (C) 2023, Advanced Micro Devices, Inc.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
@@ -27,6 +28,10 @@
 #include <xil_cache.h>
 #include <xil_exception.h>
 #include <xparameters.h>
+
+#ifdef SDT
+#include <metal/system/generic/xlnx/sys_devicetree.h>
+#endif
 
 #ifndef __METAL_GENERIC_XLNX_SYS__H__
 #define __METAL_GENERIC_XLNX_SYS__H__
--- a/lib/system/generic/xlnx/sys_devicetree.h
+++ b/lib/system/generic/xlnx/sys_devicetree.h
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) 2022-2023 Advanced Micro Devices, Inc. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
+ * @file	generic/xlnx/sys_devicetree.h
+ * @brief	Xilinx-AMD Specific symbols for System Device Tree Flow
+ */
+
+#ifndef __METAL_GENERIC_SYS_XLNX_SYS_DEVICETREE_H__
+#define __METAL_GENERIC_SYS_XLNX_SYS_DEVICETREE_H__
+
+/* The Xilinx-AMD System Device Tree (SDT) Flow BSP provides this file. */
+#include "bspconfig.h"
+
+#ifndef XPAR_SCUGIC_0_DEVICE_ID
+#define XPAR_SCUGIC_0_DEVICE_ID XPAR_SCUGIC_SINGLE_DEVICE_ID
+#endif
+
+#ifndef XPAR_SCUGIC_0_DIST_BASEADDR
+#define XPAR_SCUGIC_0_DIST_BASEADDR XPAR_SCUGIC_DIST_BASEADDR
+#endif
+#endif /* __METAL_GENERIC_SYS_XLNX_SYS_DEVICETREE_H__ */


bentheredonethat:
also @arnopo  please let me know if the demo patch is OK to take... I know there are ongoing discussions about Libmetal/OpenAMP repo demos and where to put them.

lib/system/generic/xlnx/sdt.h
```diff
@@ -0,0 +1,27 @@
...
+/*
+ * @file	generic/xlnx/sdt.h
+ * @brief	Xilinx-AMD Specific symbols for System Device Tree Flow
+ */
+
+#ifndef __METAL_GENERIC_SYS_XLNX_SDT_H__
+#define __METAL_GENERIC_SYS_XLNX_SDT_H__
+
+/* The Xilinx-AMD System Device Tree (SDT) Flow BSP provides this file. */
+#ifdef SDT
```
arnopo, line 16: needed? you include this file only if `SDT` is defined

lib/system/generic/xlnx/sdt.h
```diff
@@ -0,0 +1,27 @@
...
+#include "bspconfig.h"
+#endif
+
+#ifdef XPAR_SCUGIC_SINGLE_DEVICE_ID
+#define XPAR_SCUGIC_0_DEVICE_ID XPAR_SCUGIC_SINGLE_DEVICE_ID
+#endif
+
+#ifdef XPAR_SCUGIC_DIST_BASEADDR
+#define XPAR_SCUGIC_0_DIST_BASEADDR XPAR_SCUGIC_DIST_BASEADDR
+#endif
```
arnopo, line 26: Please tell me if I'm wrong but seems here that your issue is that XPAR_SCUGIC_0_DEVICE_ID and XPAR_SCUGIC_0_DIST_BASEADDR are not defined.
in such case you perhaps the code should be:
```diff
- #ifdef XPAR_SCUGIC_SINGLE_DEVICE_ID
+ #ifdef XPAR_SCUGIC_0_DEVICE_ID
#define XPAR_SCUGIC_0_DEVICE_ID XPAR_SCUGIC_SINGLE_DEVICE_ID
#endif

- #ifdef XPAR_SCUGIC_DIST_BASEADDR
+ #ifndef XPAR_SCUGIC_0_DIST_BASEADDR
#define XPAR_SCUGIC_0_DIST_BASEADDR XPAR_SCUGIC_DIST_BASEADDR
#endif
```

Then I assume that the `0` in the definition is the remoteproc instance.  What's happen if you need more instances?

lib/system/generic/xlnx/CMakeLists.txt
```diff
@@ -8,4 +8,10 @@ collect (PROJECT_LIB_HEADERS sys.h)
 
 collect (PROJECT_LIB_SOURCES irq.c)
 
+include(CheckSymbolExists)
+check_symbol_exists(SDT "bspconfig.h" HAS_SDT)
+if (HAS_SDT)
```
arnopo, line 13: Could you provide a more explicite name for file and defines?
replace `sdt.h` by `sys_devicetree.h` ( or something similar)
replace `SDT by SYSTEM_DT`
replace `HAS_SDT by HAS_SYSTEM_DT`

lib/system/generic/xlnx/CMakeLists.txt
```diff
@@ -8,4 +8,10 @@ collect (PROJECT_LIB_HEADERS sys.h)
...
+  collect (PROJECT_LIB_HEADERS sdt.h)
+endif()
+
```
arnopo, line 16: could you remove the remaining  `# vim ...` line above ( can be done in this commit)

lib/system/generic/xlnx/CMakeLists.txt
bentheredonethat: SDT is the symbol provided in the new BSP

the other 2 can do

lib/system/generic/xlnx/CMakeLists.txt
bentheredonethat: will do

lib/system/generic/xlnx/sdt.h
bentheredonethat: correct - this patch includes the bspconfig.h and provides symbols if system device tree bsp is included. 
we iner this based on the symbol SDT being present.

lib/system/generic/xlnx/sdt.h
bentheredonethat: I am not sure what this comment means, can you explain?

I can re-work to instead be:
`
#ifndef XPAR_SCUGIC_0_DEVICE_ID
#define XPAR_SCUGIC_0_DEVICE_ID XPAR_SCUGIC_SINGLE_DEVICE_ID
#endif
#ifndef XPAR_SCUGIC_0_DIST_BASEADDR
#define XPAR_SCUGIC_0_DIST_BASEADDR XPAR_SCUGIC_DIST_BASEADDR
#endif
`

is that ok?

lib/system/generic/xlnx/CMakeLists.txt
bentheredonethat: > Could you provide a more explicite name for file and defines? replace `sdt.h` by `sys_devicetree.h` ( or something similar) replace `SDT by SYSTEM_DT` replace `HAS_SDT by HAS_SYSTEM_DT`

I replaced sdt.h and HAS_SDT as suggested.
I kept SDT symbol as this is provided by a header we don't control.

lib/system/generic/xlnx/sdt.h
arnopo: my comment concatenates 2 points:
1) use the `ifndef XPAR_SCUGIC_0_DEVICE_ID `and` ifndef XPAR_SCUGIC_0_DIST_BASEADDR`
  you answer to it. 
2) The second point is probably more a concern/question. What does the `0` mean in `XPAR_SCUGIC_0_DEVICE_ID`? It looks to me that it represents the remoteproc instance.
In such case my question is are you able to manage 2 remote processors in your system DT?
If yes you would have to define also XPAR_SCUGIC_1_DEVICE_ID and XPAR_SCUGIC_1_DIST_BASEADDR, right?

bentheredonethat: Thank you for clarification!
1 - will do.
2 - This is due to our internal-tooling's dynamic allocation of symbols. For example in our tooling a user could in theory add X GIC instances to their design. In this case the default, first GIC instance 0-based is IP_0, IP_1 ... and so on for further instances of any IP added.

Does that make sense? To clarify these are other symbols that the BSP provides that we don't control.

bentheredonethat: > my comment concatenates 2 points:
> 
> 1. use the `ifndef XPAR_SCUGIC_0_DEVICE_ID `and` ifndef XPAR_SCUGIC_0_DIST_BASEADDR`
>    you answer to it.
> 2. The second point is probably more a concern/question. What does the `0` mean in `XPAR_SCUGIC_0_DEVICE_ID`? It looks to me that it represents the remoteproc instance.
>    In such case my question is are you able to manage 2 remote processors in your system DT?
>    If yes you would have to define also XPAR_SCUGIC_1_DEVICE_ID and XPAR_SCUGIC_1_DIST_BASEADDR, right?

updated with point 1's  #ifndef's as suggested. confirmed compile step too with this.

arnopo: > Thank you for clarification! 1 - will do. 2 - This is due to our internal-tooling's dynamic allocation of symbols. For example in our tooling a user could in theory add X GIC instances to their design. In this case the default, first GIC instance 0-based is IP_0, IP_1 ... and so on for further instances of any IP added.
> 
> Does that make sense? To clarify these are other symbols that the BSP provides that we don't control.

My concern behind my comment is that this file may not be reliable, as it depends on the system DT generation that, as you said, you do not control.
I would like to be sure that we do not received PR from users of your tool to add new definitions for each instance of IP added.

bentheredonethat: That is a valid concern. A Few things for context:

- Today AMD-Xilinx has to support both the classic flow (original xlnx/sys.h) and new SDT flow (xlnx/sys_devicetree.h). Eventually the old way will be deprecated and at that point we will fold in the xlnx/sys_devicetree.h into the xlnx/sys.h. This means the API's sys_irq_enable and sys_irq_disable in xlnx/sys.h may have to account for multiple GIC instances in the future regardless.

- If a user were to add multiple GIC instances and make them mutable from this target processor we would have to extend the API's sys_irq_enable and sys_irq_disable to specify the GIC instance too - so that at least provides a path forward if we did have to account for such a problem

- Adding GIc instances is not to my knowledge a common value-add proposition so it might be overkill to consider that corern-case since a user can use 1 GIC instance to map interrupts on a large number of managed cores.

bentheredonethat: @arnopo  does this address your concerns?

arnopo: Thanks for the additional information. 
Let's keep it like this for the moment and see how the system DT will impact libmetal in the future

arnopo (approved):

PR #282, opened 2024-01-15
lib: xlnx: Avoid mapping in MPU for region mapped by bsp

tnmysh:
Avoiding memory mapping in MPU config table for region that are already mapped by bsp and no change in attribute property.

### DIFF
--- a/lib/system/generic/xlnx/sys.c
+++ b/lib/system/generic/xlnx/sys.c
@@ -1,6 +1,6 @@
 /*
  * Copyright (c) 2022, Xilinx Inc. and Contributors. All rights reserved.
- * Copyright (c) 2022-2023 Advanced Micro Devices, Inc. All Rights Reserved.
+ * Copyright (c) 2022-2024 Advanced Micro Devices, Inc. All Rights Reserved.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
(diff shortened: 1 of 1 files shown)


tnmysh:
Changes in v2:
  - Remove CamelCasing for variables defined in libmetal code.
  - Fix spacing related warnings.

tnmysh:
@arnopo, @edmooring for reviews.

tnmysh:
This is workaround while this issue is resolved: https://github.com/OpenAMP/libmetal/issues/283

tnmysh:
> As we discussed during the OpenAMP call, the current code can serve as a temporary workaround.
> However, in the long term, for the sys.c, the objective is file to act as a simple wrapper that calls an AMD system HAL outside of libmetal. This AMD HAL should address this kind of specificity.

Ack.

tnmysh:
Updated correct author signature.

tnmysh:
@edmooring ping for reviews.

lib/system/generic/xlnx/sys.c
```diff
@@ -72,12 +72,63 @@ void metal_weak metal_generic_default_poll(void)
...
+	u32 req_addr = pa;
+	u32 mmap_req = 1;
+	u32 base_end_addr;
+	u32 cnt;
+
+	/* Get the MPU Config enties */
+	Xil_GetMPUConfig(mpu_config);
+
+	for (cnt = 0; (cnt < MAX_POSSIBLE_MPU_REGS) &&
+	     (mpu_config[cnt].flags & XMPU_VALID_REGION); cnt++) {
```
edmooring, line 95: This test doesn't seem quite right. If the MPU registers are sparse, for example 0,1 and 3 have valid regions, but 2 does not, then 3 will be skipped in this loop. It's been quite awhile since I looked at the MPU specs, but I think the hardware allows this.

Maybe this test should be moved inside the loop, with an if() continue statement?

edmooring (changes requested):
Otherwise, looks good for now.

lib/system/generic/xlnx/sys.c
tnmysh: Thanks @edmooring It's done.

tnmysh:
@edmooring addressed comment as requested. Ping for reviews.

edmooring (approved):
Looks good to go.

PR #284, opened 2024-02-01
FreeRTOS Avoid mem mapping in MPU for region mapped

rajimoha:
FreeRTOS patch for Avoiding memory mapping in MPU config table for region that are already mapped by bsp and no change in attribute property

### DIFF
--- a/lib/system/freertos/xlnx/sys.c
+++ b/lib/system/freertos/xlnx/sys.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023, Advanced Micro Devices, Inc.
+ * Copyright (C) 2024, Advanced Micro Devices, Inc.
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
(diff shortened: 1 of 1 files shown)


lib/system/freertos/xlnx/sys.c
```diff
@@ -75,12 +75,64 @@ void metal_weak metal_generic_default_poll(void)
...
+	u32 req_addr = pa;
+	u32 mmap_req = 1;
+	u32 base_end_addr;
+	u32 cnt;
+
+	/* Get the MPU Config enties */
+	Xil_GetMPUConfig(mpu_config);
+
+	for (cnt = 0; (cnt < MAX_POSSIBLE_MPU_REGS) &&
+			(mpu_config[cnt].flags & XMPU_VALID_REGION); cnt++){
```
edmooring, line 100: This loop will terminate early if the MPU register mappings are sparse. If registers 0, 1, and 3 have valid regions, but 2 does not, 3 will be skipped. 

I think this part of the test should be moved into the loop as an if() continue statement.

edmooring (changes requested):
Need to fix the formatting issues as well as address my comment.

rajimoha:
@edmooring changes done.. ping for review

edmooring (approved):
Looks good to go.

arnopo (changes requested):
commit message format to fix

arnopo (approved):

PR #294, opened 2024-05-03
CI: fix timezone and pip3 issues

arnopo:
The build check fails with following errors:

1) ln: failed to create symbolic link '/etc/localtime': File exists
Fix it by removing the link creation

2) error: externally-managed-environment
Fix it by adding "--break-system-packages" option to pip3

### DIFF
--- a/.github/actions/build_ci/Dockerfile
+++ b/.github/actions/build_ci/Dockerfile
@@ -1,10 +1,10 @@
-FROM ubuntu:latest
+FROM ubuntu:24.04
 
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 
 # Install prerequisites
-RUN apt-get --quiet=2 update && apt-get install --quiet=2 --assume-yes sudo git python3 python3-pip wget
+RUN apt-get --quiet=2 update && apt-get install --quiet=2 --assume-yes sudo git python3 python3-pip python3-venv wget
 
 # Copies your code file from your action repository to the filesystem path `/` of the container
 COPY entrypoint.sh /entrypoint.sh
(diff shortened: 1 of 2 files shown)


tnmysh (commented):
LGTM.

wmamills:
Please wait until we agree on the way forward for open-amp and then we/I can update this one also.

wmamills:
@arnopo are you going to do this one or do you want me to?  I had it started but I don't have the change to use venv

arnopo:
> @arnopo are you going to do this one or do you want me to? I had it started but I don't have the change to use venv

Both are ok for me. I'm waiting that we fix open-amp. I did not start libmetal update yet.

arnopo:
I updated to use ubuntu 24.04 environement
I also fix dead link for the freeRTOS lib
the checkpatch error is due to the update of the FreeRTOS URL

wmamills:
I pulled this down and pushed to my main branch to test on github.
I get a failure on zephyr.
I also see that Linux gets the same failures I was seeing on the desktop testing but github does not count them as an error.
My job log is here:
https://github.com/wmamills/libmetal/actions/runs/9004301644/job/24736958627

wmamills:
The zephyr failure appears to be a temp failure of the github web server.  I ran the same thing again and it worked OK for zephyr.
The linux test failures are still there.  I switched to 22.04 for the docker image (on desktop) and they are still there.

wmamills:
The zephyr failures are hard to figure out.  Initially they failed on the desktop, now they pass with me changing anything.  They still fail on github but this does not cause a problem right now as github workflow ignores the error.

I can't spend any more time on this right now so I suggest accepting this PR as is and we can fixup the Linux failures later (or delete them as they are redundant with the other test action).

FYI: I tested this commit to fix github ignoring the test results.  It does allow the workflow to fail but I don't have a fix or explanation of the failure right now.
https://github.com/wmamills/libmetal/actions/runs/9020751918

arnopo:
> The Linux failures are hard to figure out. Initially they failed on the desktop, now they pass without me changing anything. They still fail on github but this does not cause a problem right now as github workflow ignores the error.
> 

Yes I suspect that the issue is there since a while. As the workflow don't test the output it ignore the issue.

> I can't spend any more time on this right now so I suggest accepting this PR as is and we can fixup the Linux failures later (or delete them as they are redundant with the other test action).

i will try to have a look this week. Ok to treat this in a separate PR/issue

> 
> FYI: I tested this commit to fix github ignoring the test results. It does allow the workflow to fail but I don't have a fix or explanation of the failure right now. https://github.com/wmamills/libmetal/actions/runs/9020751918

PR #302, opened 2024-07-16
Xlnx cmake machine name change

bentheredonethat:
Note that the second commit is just for consistency and can be removed -- It is known that the apps are moving to separate repo.

The intent here is to make the xlnx section of libmetal lib to be SOC agnostic.
CC @tnmysh  @arnopo @wmamills


arnopo (commented):
should `zynqmp_amp_demo` become` amp_demo`?

bentheredonethat:
@arnopo  yes that makes sense. The patch of more interest from my perspective is the 1/2 patch.

bentheredonethat:
@arnopo  updated.

lib/system/generic/CMakeLists.txt
```diff
@@ -17,10 +17,10 @@ collect (PROJECT_LIB_SOURCES irq.c)
...
 collect (PROJECT_LIB_SOURCES time.c)
 
-if ("${PROJECT_MACHINE}" STREQUAL "zynqmp_a53" OR
-    "${PROJECT_MACHINE}" STREQUAL "zynqmp_a72" OR
-    "${PROJECT_MACHINE}" STREQUAL "zynqmp_a78" OR
-    "${PROJECT_MACHINE}" STREQUAL "zynqmp_r5"  OR
+if ("${PROJECT_MACHINE}" STREQUAL "xlnx_a53" OR
+    "${PROJECT_MACHINE}" STREQUAL "xlnx_a72" OR
+    "${PROJECT_MACHINE}" STREQUAL "xlnx_a78" OR
+    "${PROJECT_MACHINE}" STREQUAL "xlnx_r5"  OR
```
tnmysh, line 23: These machine names need to be chosen carefully.
So, Mulit-core platforms like zynqmp, has different configuration in same core. I can imagine three machine names in zynqmp: R5-lockstep-core, R5-split-core0 and R5-split-core1.

If we go that route, we might end up, with long list of MACHINE names, as future plaforms might have more such cores and configurations.

@arnopo , @wmamills any other approach in such case ?

tnmysh: Or, we can introduce other cmake variables such as:

PROJECT_MACHINE=zynqmp, PROJECT_CORE=a53/R5, PROJECT_CORE_CONF=lockstep/split, PROJECT_CORE_NUMBER="0, 1 etc... ?

bentheredonethat: Hi @tnmysh  from Xilinx-amd side, the core number and SOC for latest workflows (SDT) is inferred via header files. 

That is 'zynqmp' / 'versal' etc and core 0/ core1 are symbols in the BSP. That being said we can update BSP to output those two. lockstep/split is not always visible from tooling in Xilinx-amd so in many cases lockstep==rpu0 for the defaults.

if it is custom tooling then we can add separate handling for lockstep but again, rpu0 defaults should already be ok for lockstep unless you mean for lockstep it should be using all the available TCM banks for that cluster. please clarify what is needed for the lockstep case in default demos.

bentheredonethat: though separately we would like to see if it is possible to remove the demos (can open separate PR / discussion for that)

tnmysh: > That is 'zynqmp' / 'versal' etc and core 0/ core1 are symbols in the BSP.

In such case, I believe we shouldn't need xlnx_a53, xlnx_a72 both.
As zynqmp -> a53, versal -> a72. So, can we just have "xlnx" Instead?
How do we distinguish between cortex-a and cortex-r ? Headers also provide such defines?

bentheredonethat: for cortex-a and cortex-r (and microblaze) we can distinguish from symbols. But if we do that for MB too then we will have to do something to account for the different logic in MB subsection.
can we have 'xlnx_a' , xlnx_r' and 'xlnx_mb' ?

bentheredonethat: but @tnmysh  for downstream consumers (openamp) how will we determine the SOC/core etc in the CMake? do you want us to pick things up from CMake checksymbolexists ?

tnmysh: Exactly, that is why I was suggesting to have more config variables, and those should be used across various software stacks i.e. apps, open-amp and libmetal.
My suggestion was for more generic case that can handle not xlnx platforms, but any multi-core platform.

May be we end up addressing it in different PR, but I am looking for suggestion that is scalable to other platforms too, not only xlnx platform.

bentheredonethat: @tnmysh   adding adding machine name for SOC/Core combo is fair since we see that combination across multiple vendor/SOC's for remoteproc as well as in open source libs.

what that means is 'versal_a72' is a common-sense convention. it has the drawback of being redundant with headers but scales for other SOC's but is sufficient for many cases. Core or vendor specific configurations (lockstep for rpu) can be captured in header symbols.

arnopo: PROJECT_MACHINE is not a generic variable and seems only used by AMD.

The objective here is to include `xlnx` subdirectory, right ?

What about a directory hierarchy `vendor/platform/core` or `vendor/core`?

a `PROJECT_VENDOR` ( or another variable) could define the vendor sub-directory and then  in vendor directory the core can be differentiated, if needed, by `PROJECT_MACHINE` or `PROJECT_TARGET`

bentheredonethat: Hi @arnopo  thanks for the comment here.
so if I understand your suggestion then the libmetal usage would be as follows:

- add new cmake var PROJECT_VENDOR
- in libmetal lib ports we can just keep the vendor (each vendor can then organize that as needed, xilinx-amd would just keep the xlnx subdir)

in cmake lib simply validate PROJECT_VENDOR and then in the vendor-specific CMake there can be additional validation of the  PROJECT_MACHINE/PROJECT_TARGET as needed

@tnmysh  @arnopo  are you OK with that proposal?

arnopo: This triggers a discussion on how to harmonize the folders in the different repositories to add more platform support. Let's add this to the agenda for the next RP call.

wmamills (changes requested):

bentheredonethat:
@wmamills  updated first commit per your review.
ill have to look at the default / non-vendor case still.

bentheredonethat:
the way it is now is keeping the pre-exissting CMake Machine logic check (template machine exists)
if you want to switch from using machine at all in the general/non-vendor case then i can say to use 'template' since we already have that.

please advise.

wmamills:
@bentheredonethat how are you calling cmake for your builds?
I was expecting cmake/platforms/zynqmp-{a53,r5}-{freertos,generic}.cmake to change or be collapsed to one per RTOS.
If it were me I would add xlnx-freertos.cmake and xlnx-generic.cmake and either delete the others or make them symlinks for back compatibility.  Don't add symlinks for a72/a73 as the don't exist today.  In the xlnx-* templates you should define the PROJECT_VENDOR.

For handling the non-vendor case I would add a default of "none" for PROJECT_VENDOR in cmake/options.cmake.  You can look at the handling of MACHINE in that file.  I would expect something like: (untested)
if (NOT DEFINED PROJECT_VENDOR)
  set (PROJECT_VENDOR "none")
endif (NOT DEFINED PROJECT_VENDOR)
message ("-- Vendor: ${PROJECT_VENDOR}")

Since the lib/freertos/none will never exist this will prevent your current logic from including lib/freertos twice.  I am not sure what that will do in cmake so best to avoid it.

bentheredonethat:
@wmamills  the PROJECT_VENDOR part i can add and test.

for the cmake/platforms part -- im taking a look at that and not sure if I understand how that would work. I took a look and the thing is we still want the PROJECT_MACHINE for the microblaze case and for demos.

additionally the CMAKE_C_FLAGS, CMAKE_SYSTEM_PROCESSOR, and CROSS_PREFIX are all different. having multiple differences per machine means it makes sense to keep them as is and (to me at least) just add the PROJECT_VENDOR to each of these.

i have proceeded with the above in latest commits.

wmamills:
@bentheredonethat yes you are right about the templates.  I looked at the new commit and the template files look OK.

We still need a default value for PROJECT_VENDOR as I outlined above as the base case fails with out it.
See the CI results for build-generic-arm.
Build Failure is: 
CMake Error at lib/system/generic/CMakeLists.txt:21 (add_subdirectory):
  add_subdirectory called with incorrect number of arguments
-- Configuring incomplete, errors occurred!

I will try what I suggested here to make sure it works.

You can run the generic build yourself, just look at the build_generic function in:
.github/actions/build_ci/entrypoint.sh
(I would add -DWITH_DOC=OFF to the cmake line so you don't need to worry about the docs stuff.  It turns itself off in the needed programs are not found but if you have some and not all it will fail.)

bentheredonethat:
tested and built as suggested above. Thanks @wmamills !

wmamills:
I tested as well and the generic-build works OK on my desk and in CI.

CI is getting an error on the Zephyr step but that looks like an upstream change not something in the PR.
I am proving that to myself now.

wmamills:
Background on Zephyr error:
https://github.com/zephyrproject-rtos/zephyr/pull/74682
Zephyr v3.7.0 or main has this PR.  I am trying v3.6.0 which is what we used in the v2024.05 release.

wmamills:
With Zephyr locked to v3.6.0 the CI passes.
https://github.com/wmamills/libmetal/actions/runs/10395182873/job/28786640624
Fixing Zephyr v3.7.0 is not an issue for this PR.

diff --git a/.github/actions/build_ci/entrypoint.sh b/.github/actions/build_ci/entrypoint.sh
index 9333c32..99eb283 100755
--- a/.github/actions/build_ci/entrypoint.sh
+++ b/.github/actions/build_ci/entrypoint.sh
@@ -5,6 +5,7 @@ readonly TARGET="$1"
 ZEPHYR_TOOLCHAIN_VARIANT=zephyr
 ZEPHYR_SDK_API_FOLDER=https://api.github.com/repos/zephyrproject-rtos/sdk-ng/releases/latest
 ZEPHYR_SDK_SETUP_TAR=zephyr-sdk-.*linux-x86_64.tar.xz
+ZEPHYR_VER=v3.6.0
 
 FREERTOS_ZIP_URL=https://sourceforge.net/projects/freertos/files/FreeRTOS/V10.0.1/FreeRTOSv10.0.1.zip
 
@@ -82,7 +83,7 @@ build_zephyr(){
        pv $ZEPHYR_SDK_TAR -i 3 -ptebr -f | tar xJ || exit 1
        rm -rf $ZEPHYR_SDK_TAR || exit 1
        yes | ./$ZEPHYR_SDK_SETUP_DIR/setup.sh || exit 1
-       west init ./zephyrproject || exit 1
+       west init --mr $ZEPHYR_VER ./zephyrproject || exit 1
        cd ./zephyrproject || exit 1
        west update --narrow || exit 1
        west zephyr-export || exit 1

wmamills (approved):
New version looks good now.

arnopo:
Hi  @bentheredonethat 
Any reason to not have same folder hiearchy  for Linux examples, using `PROJECT_VENDOR`?

bentheredonethat:
@arnopo  thanks for the comment. in that case would it be:
examples -> xlnx -> xlnx_r5
?

arnopo:
> @arnopo thanks for the comment. in that case would it be: examples -> xlnx -> xlnx_r5 ?

The LInux example does not run on the R5 but on the A53, right?
what about `examples/system/linux/xlnx/zynqmp_amp_demo/` -> `examples/system/linux/xlnx_a53/amp_demo/`

bentheredonethat:
> > @arnopo thanks for the comment. in that case would it be: examples -> xlnx -> xlnx_r5 ?
> 
> The LInux example does not run on the R5 but on the A53, right? what about `examples/system/linux/xlnx/zynqmp_amp_demo/` -> `examples/system/linux/xlnx_a53/amp_demo/`

Should this not be:
> The LInux example does not run on the R5 but on the A53, right? what about > `examples/system/linux/xlnx/zynqmp_amp_demo/` -> `examples/system/linux/xlnx/xlnx_a53/amp_demo/`

arnopo:
> > > @arnopo thanks for the comment. in that case would it be: examples -> xlnx -> xlnx_r5 ?
> > 
> > 
> > The LInux example does not run on the R5 but on the A53, right? what about `examples/system/linux/xlnx/zynqmp_amp_demo/` -> `examples/system/linux/xlnx_a53/amp_demo/`
> 
> Should this not be:
> 
> > The LInux example does not run on the R5 but on the A53, right? what about > `examples/system/linux/xlnx/zynqmp_amp_demo/` -> `examples/system/linux/xlnx/xlnx_a53/amp_demo/`


Reading your PR again , I think that my advice is not accurate.  What I would like here for the folder hierachy to be coherent for the whole project.
I would expect having the same hierarchy in lib and examples following the template `examples/system/${PROJECT_SYSTEM}/${PROJECT_VENDOR} ` 

So 
`examples/system/linux/xlnx/zynqmp_amp_demo/` -> `examples/system/linux/xlnx/xlnx_a53/amp_demo/`
`examples/system/generic/xlnx_r5/amp_demo/` -> `examples/system/generic/xlnx/amp_demo/`
`examples/system/freertos/xlnx_r5/amp_demo/` -> `examples/system/freertos/xlnx/amp_demo/`

bentheredonethat:
@arnopo  I am good with that. will send patch as suggested above. thanks

Issue #295, opened 2024-05-05
CI: Pin to specific version of Zephyr SDK in default PRs


wmamills:
Unlike open-amp CI, this script uses the very latest version of the Zephyr SDK on each PR check.
This is bad.  While it is good to test against the latest SDK, that should not be mixed into checking a PR.

The default PR should use a pinned version of the SDK (and the same one as open-amp).

It would be good to have separate CI jobs that would test a known good library (current main should be fine) against the latest Zephyr SDK, latest Zephyr, latest Ubuntu etc

wmamills:
@arnopo what do you want to do about this?  Right now libmetal is using 0.16.6 for CI and can change at any time.  open-amp is using 0.16.1 (which is old but at least a fixed version.)

wmamills:
This issue caused problems again in PR #302 .
If CI randomly fails for reasons unrelated to the CI, it just teaches people to ignore the CI results.
Alternatively if Upstream Zephyr has a breaking change and we have no PRs, we don't know about it until the next PR.

Testing libmetal against latest Zephyr upstream should be its own job and should be run weekly using libmetal main.
Testing in PRs and merges should be against a known good version of Zephyr and Zephyr SDK.

arnopo:
> This issue caused problems again in PR #302 . If CI randomly fails for reasons unrelated to the CI, it just teaches people to ignore the CI results. Alternatively if Upstream Zephyr has a breaking change and we have no PRs, we don't know about it until the next PR.
> 
> Testing libmetal against latest Zephyr upstream should be its own job and should be run weekly using libmetal main. Testing in 

Seems reasonable. What I cannot understand is why it fails with Zephyr 3.7.99 as this is the last release of zephyr.

PR #307, opened 2024-08-18
CI: Do Zephyr build tests on known good and latest versions

wmamills:
Allow Zephyr testing to either use locked known good values or the very latest versions.

The known good versions are best for PR checking as if the build fails it is almost always the PR itself that broke it.

The latest version is good for periodically checking compatibility with the very latest Zephyr changes.

For now we run both on pushes and PRs.

We also run main against the latest zephyr check weekly as a look ahead.

Fixes: #295

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,7 +22,7 @@ jobs:
     name: check builds on different platforms
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: build for Linux
       id: build_linux
       uses: ./.github/actions/build_ci
@@ -38,7 +38,16 @@ jobs:
       uses: ./.github/actions/build_ci
       with:
         target: freertos
-    - name: build for Zephyr
+
+  # Break the zephyr builds into their own job as the common runner was
+  # running out of space when runs were together
+  # Also, as the longest running jobs, this allows them to run in ||
+  zephyr_build_known_good_version:
+    name: Zephyr build with a version that is known to work
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: build for Zephyr (Known Good)
       id: build_Zephyr
       uses: ./.github/actions/build_ci
       with:
@@ -48,7 +57,7 @@ jobs:
     name: nonreg tests
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: configure
       run: |
         sudo apt-get update && sudo apt-get install libsysfs-dev
--- a/.github/workflows/heathcheck.yml
+++ b/.github/workflows/heathcheck.yml
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Linaro Limited.
+
+# The focus of this flow is to check for external changes that will effect
+# the project.  Even if no code changes are happening, changes in external
+# distros or projects can invalidate our work and this flow lets us know
+
+name: libmetal Heath Check
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+    paths-ignore:
+      - docs/**
+
+  # Allows you to run this workflow manually from the Actions tab or gh API
+  workflow_dispatch:
+
+  # run weekly on Sunday at 5:10 AM UTC (9:10 PM US western)
+  schedule:
+    - cron:  '10 5 * * 0'
+
+jobs:
+  zephyr_build_main:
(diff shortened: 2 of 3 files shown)


wmamills:
@arnopo 
This branch does not contain the v3.7.0 fix so the "latest" builds will always fail.
To show the logic I have hard coded Zephyr v3.6.0 and SDK 0.16.5-1 so they will be noticabbly different than latest.
We should merge the fix for v3.7.0 before this one and I will rebase and then set known good values of v3.7.0 and 0.16.8

wmamills:
@arnopo @tnmysh @edmooring 
rebased to main.
Now that main is compatible with Zephyr v3.7.0, use that as the known good version.
Also use 0.16.8 as the known good SDK version (which will also be the latest for now)
The above is what we should see in a PR.  If a PR fails against Zephr known good it should not be merged.
If a PR fails the Heath Check Zephyr latest, that is probably not something the PR.  You can check by checing the latest main heath check or running a new one manualy via the web interface.

.github/workflows/heathcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+  workflow_dispatch:
+
+  # run weekly on Sunday at 5:10 AM UTC (9:10 PM US western)
+  schedule:
+    - cron:  '10 5 * * 0'
+
+jobs:
+  zephyr_build_latest:
+    name: 'Zephyr Latest build'
+    runs-on: ubuntu-latest
```
arnopo, line 28: the "lastest" term seems quite obscure to me as it can represent the lastest release.
what about "zephyr_main_build"

.github/workflows/continuous-integration.yml
```diff
@@ -38,7 +38,16 @@ jobs:
       uses: ./.github/actions/build_ci
       with:
         target: freertos
-    - name: build for Zephyr
+
+  # Break the zephyr builds into their own job as the common runner was
+  # running out of space when runs were together
+  # Also, as the longest running jobs, this allows them to run in ||
+  zephyr_build_known_good:
```
arnopo, line 45: what about stable instead of "good" ?

.github/workflows/heathcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+    branches: [ main ]
+    paths-ignore:
+      - docs/**
+
+  # Allows you to run this workflow manually from the Actions tab or gh API
+  workflow_dispatch:
+
+  # run weekly on Sunday at 5:10 AM UTC (9:10 PM US western)
+  schedule:
+    - cron:  '10 5 * * 0'
```
arnopo, line 23: question: what happen if it fails. A message is sent on the mailing list, or do we have to check action result each week?

.github/workflows/continuous-integration.yml
tnmysh: Ack.

wmamills: I really don't like stable.  The know good version does not need to be a release, just a recent point that is know to work.  If it is a release, it does not need to be the most recent stable release.

However "know good version" is a term I have used for decades and I am not the only one:
https://github.com/GoogleChromeLabs/chrome-for-testing?tab=readme-ov-file#json-api-endpoints

I can add _version.  That makes it easier to understand.

wmamills:
Pushed with some language updates as describe above.

wmamills:
> question: what happen if it fails. A message is sent on the mailing list, or do we have to check action result each week?
The answer seem to be here:
https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/notifications-for-workflow-runs

I don't see anyway to send that to an email list.  We can change who it goes to by disable and then enable.  The last enable is the winner.  

We could also add a badge on the readme.

arnopo:
> @arnopo why do we ignore changes to cmake/* and scripts/*?? Changes there are the most likely to cause problems for the builds.

Mainly to avoid issue with checkpatch and to avoid running build on non-library code (no true for cmake)

> 
> Are you OK if I start another PR to remove those ignore paths?

yes sure!

arnopo:
> > question: what happen if it fails. A message is sent on the mailing list, or do we have to check action result each week?
> 
> The answer seem to be here: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/notifications-for-workflow-runs
> 
> I don't see anyway to send that to an email list. We can change who it goes to by disable and then enable. The last enable is the winner.
> 
> We could also add a badge on the readme.

Regarding the article you pointed out and this article: https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28,
we should be able to return another status such as `neutral `, or `action_required`

In such case we could run it for each PR but returning something else than `failure` with an explicit print in action log:
" If the error reported is not related to your PR, you can ignore it, it probably related to an internal Zephyr error"

arnopo (approved):

PR #315, opened 2024-10-10
lib/system/nuttx: fix undeclared parameter compile error

wyr-7:
fix compile error:
/home/wyr/work/code/project/cardev/nuttx/openamp/libmetal/lib/system/nuttx/device.c:16:22: error: 'io' undeclared (first use in this function)
   16 |         metal_unused(io);
/home/wyr/work/code/project/cardev/nuttx/openamp/libmetal/lib/system/nuttx/device.c:14:53: error: unused parameter 'dev' [-Werror=unused-parameter]
   14 | int metal_generic_dev_sys_open(struct metal_device *dev)
...

### DIFF
--- a/lib/system/nuttx/device.c
+++ b/lib/system/nuttx/device.c
@@ -13,5 +13,7 @@
 
 int metal_generic_dev_sys_open(struct metal_device *dev)
 {
+	metal_unused(dev);
+
 	return 0;
 }
--- a/lib/system/nuttx/init.c
+++ b/lib/system/nuttx/init.c
@@ -18,6 +18,7 @@ struct metal_state _metal;
 int metal_sys_init(const struct metal_init_params *params)
 {
 	int ret = metal_cntr_irq_init();
+	metal_unused(params);
 
 	if (ret >= 0)
 		ret = metal_bus_register(&metal_generic_bus);
--- a/lib/system/nuttx/io.c
+++ b/lib/system/nuttx/io.c
@@ -14,6 +14,7 @@ static uint64_t metal_io_read_(struct metal_io_region *io,
 			       int width)
 {
 	uint64_t value = 0;
+	metal_unused(order);
 
 	metal_io_block_read(io, offset, &value, width);
 	return value;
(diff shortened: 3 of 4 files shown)


lib/system/nuttx/device.c
```diff
@@ -13,5 +13,6 @@
 
 int metal_generic_dev_sys_open(struct metal_device *dev)
 {
+	metal_unused(dev);
```
arnopo, line 16: add space after

lib/system/nuttx/init.c
```diff
@@ -19,6 +19,7 @@ int metal_sys_init(const struct metal_init_params *params)
 {
 	int ret = metal_cntr_irq_init();
 
+	metal_unused(params);
```
arnopo, line 22: Nitpicking: To respect the coding rules used in other source code, consider declaring metal_unused as a variable at the top of the function (to apply in other functions as well).

Here it becomes:
```
int metal_sys_init(const struct metal_init_params *params)
{
	int ret = metal_cntr_irq_init();
	metal_unused(params);
	
	if (ret >= 0)
```

arnopo (changes requested):
just code style remarks

tnmysh (approved):
LGTM after addressing arnaud's concerns.

edmooring (approved):
Looks good to go if Arnaud's suggestions are implemented.

lib/system/nuttx/device.c
wyr-7: Ok, done.

lib/system/nuttx/init.c
wyr-7: Ok, done.

wyr-7:
@arnopo Thanks, I have update. @edmooring @tnmysh

arnopo (approved):
Nit: We usually use the same approach as in the Linux Kernel for the hierarchy description in the commit subject. I will update the commit subject for this one. Please take it into account for other PRs.

`lib/system/nuttx: fix undeclared parameter compile error`
should become
`lib: system; nuttx: fix undeclared parameter compile error`

PR #316, opened 2024-10-10
libmetal(cmake): set HAVE_STDATOMIC_H default true in NuttX platform

wyr-7:
there is no need to use cmake include detection under NuttX platform

### DIFF
--- a/cmake/depends.cmake
+++ b/cmake/depends.cmake
@@ -24,7 +24,7 @@ if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
   find_package(LibRt REQUIRED)
   collect (PROJECT_LIB_DEPS "${LIBRT_LIBRARIES}")
 
-else ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+elseif (NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "NuttX")
 
   # TODO: fix for find_path() to detect stdatomic.h
   # find_path (HAVE_STDATOMIC_H stdatomic.h)
@@ -33,5 +33,5 @@ else ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
   check_include_files (stdatomic.h HAVE_STDATOMIC_H)
   set (CMAKE_REQUIRED_FLAGS ${_saved_cmake_required_flags})
 
-endif ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+endif ()
 


cmake/depends.cmake
```diff
@@ -24,7 +24,13 @@ if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
   find_package(LibRt REQUIRED)
   collect (PROJECT_LIB_DEPS "${LIBRT_LIBRARIES}")
 
-else ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+elseif ("${CMAKE_SYSTEM_NAME}" STREQUAL "NuttX")
+
+  # there is no need to use cmake include detection
+  # under NuttX platform
+  set(HAVE_STDATOMIC_H true)
```
arnopo, line 31: Seems not a good idea to force `HAVE_STDATOMIC_H` here
what about ujsrt having 
```
elseif (NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "NuttX")

  # TODO: fix for find_path() to detect stdatomic.h
  # find_path (HAVE_STDATOMIC_H stdatomic.h)
  set (_saved_cmake_required_flags ${CMAKE_REQUIRED_FLAGS})
  set (CMAKE_REQUIRED_FLAGS "-c" CACHE STRING "")
  check_include_files (stdatomic.h HAVE_STDATOMIC_H)
  set (CMAKE_REQUIRED_FLAGS ${_saved_cmake_required_flags})

endif ()
```

xuxin930: NuttX platform supports builtin atomic and toolchain atomic.
The reason why CMAKE_REQUIRED_FLAGS cannot be used here is because this cache will destroy the CMake build of the integrated platform

arnopo: My point here is that `HAVE_STDATOMIC_H ` is probably defined in NuttX  so the   `set(HAVE_STDATOMIC_H true)` seems redundant. 
Exclude the cmake code related to `CMAKE_REQUIRED_FLAGS` could be sufficient

cmake/depends.cmake
```diff
@@ -24,7 +24,13 @@ if ("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
...
+
+else ()
```
xuxin930, line 33:
```suggestion
elseif (NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "NuttX")
```

xuxin930: CC @arnopo @wyr-7

wyr-7:
@arnopo Thansk, I have updated it according to @xuxin930 comment.

arnopo (approved):
LGTM

edmooring (approved):
Looks good to go.

arnopo:
I have fixed the commit subject that was not up to date

PR #317, opened 2024-10-10
lib/system/nuttx/io.c: include <stddef.h> in nuttx/io.c

wyr-7:
Because nuttx/io.c use NULL and NULL is defined in <stddef.h>


xiaoxiang781216 (approved):

arnopo (approved):

edmooring (approved):
Looks good to go.

PR #319, opened 2024-10-24
lib: processor: hosted: fix compilation warning in metal_cpu_yield

arnopo:
Fix cross inclusion between metal/sys.h and /processor/hosted/cpu.h replacing metal_wait_usec by k_busy_wait function call.

Issue reported by Zephyr twister CI test: https://github.com/zephyrproject-rtos/zephyr/actions/runs/11482800628/job/31962168738?pr=80324

libmetal/lib/include/metal/processor/hosted/cpu.h:19:9: error: implicit declaration of function ‘metal_wait_usec’
   19 |         metal_wait_usec(10);
      |         ^~~~~~~~~~~~~~~
libmetal/lib/include/metal/system/zephyr/sys.h:46:20:
error: conflicting types for ‘metal_wait_usec’; have ‘void(uint32_t)’
   46 | static inline void metal_wait_usec(uint32_t usec_to_wait)
      |                    ^~~~~~~~~~~~~~~
libmetal/lib/include/metal/processor/hosted/cpu.h:19:9:
note: previous implicit declaration of ‘metal_wait_usec’ with
type ‘void(uint32_t)’ {aka ‘void(unsigned int)’}
   19 |         metal_wait_usec(10);
      |         ^~~~~~~~~~~~~~~

### DIFF
--- a/lib/processor/hosted/cpu.h
+++ b/lib/processor/hosted/cpu.h
@@ -9,14 +9,9 @@
  * @brief	Hosted environment CPU specific primitives
  */
 
-#include <metal/sys.h>
-
 #ifndef __METAL_HOSTED_CPU__H__
 #define __METAL_HOSTED_CPU__H__
 
-static inline void metal_cpu_yield(void)
-{
-	metal_wait_usec(10);
-}
+void metal_cpu_yield(void);
 
 #endif /* __METAL_HOSTED_CPU__H__ */
--- a/lib/processor/CMakeLists.txt
+++ b/lib/processor/CMakeLists.txt
@@ -5,10 +5,13 @@ else()
 	collect (PROJECT_LIB_HEADERS generic/cpu.h)
 endif()
 
+if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_PROCESSOR}/cpu.c")
+	collect (PROJECT_LIB_SOURCES ${PROJECT_PROCESSOR}/cpu.c)
+endif()
+
 if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_PROCESSOR}/atomic.h")
 	collect (PROJECT_LIB_HEADERS ${PROJECT_PROCESSOR}/atomic.h)
 	set(HAVE_PROCESSOR_ATOMIC_H 1 CACHE INTERNAL "Have include ${PROJECT_PROCESSOR}/atomic.h")
 else()
 	collect (PROJECT_LIB_HEADERS generic/atomic.h)
 endif()
-
--- a/lib/processor/hosted/cpu.c
+++ b/lib/processor/hosted/cpu.c
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) 2024, STMicroelectronics.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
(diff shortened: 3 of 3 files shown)


lib/processor/hosted/cpu.h
```diff
@@ -9,14 +9,14 @@
...
-#include <metal/sys.h>
+#include <zephyr/kernel.h>
 
 #ifndef __METAL_HOSTED_CPU__H__
 #define __METAL_HOSTED_CPU__H__
 
 static inline void metal_cpu_yield(void)
 {
-	metal_wait_usec(10);
+	k_busy_wait(10);
```
tnmysh, line 19: This file is processor specific. k_busy_wait is for zephy OS.
 Should this API not included in zephyr specific header?
May be I am missing something, but I don't understand how metal_cpu_yield will be implemented for FreeRTOS, or baremetal fw of different platforms.

lib/processor/hosted/cpu.h
```diff
@@ -16,7 +16,7 @@
 
 static inline void metal_cpu_yield(void)
 {
-	metal_wait_usec(10);
+		k_busy_wait(10);
```
tnmysh, line 19: This should be removed ?

tnmysh: Or I think we should update metal_wait_usec with k_busy_wait for zephyr specific case.

lib/processor/hosted/cpu.h
tnmysh: Thanks! now change looks good.

tnmysh (approved):
LGTM.

edmooring (approved):
Looks good to go.

arnopo:
Unfortunately use `k_yield` generates a regression in some zephyr CI test.  I need to find another way to solve this issue.

arnopo:
A new version of the fix involves creating a cpu.c file to avoid the inline declaration. The Zephyr CI passes with this solution 
and the implementation should preserve the legacy behavior for zephyr.

lib/processor/hosted/cpu.h
wmamills: This specific call was requested by Zephyr.
They don't want k_yield on the posix environment as it will use 100% of at least 1 cpu.
We don't need 100% of a 4GHz CPU to emulate the actions of an MCU.

"hosted" is supposed to be used when your running on your desktop/laptop.  In Zephyr it is the "posix" board or the equivelent for Windows.  If FreeRTOS has a posix mode why could they not implement metal_wait_usec???

wmamills: It looks like the new version keeps the metal_wait_usec and just moves it.  So this is fine.

arnopo: > This specific call was requested by Zephyr. They don't want k_yield on the posix environment as it will use 100% of at least 1 cpu. We don't need 100% of a 4GHz CPU to emulate the actions of an MCU.
> 
> "hosted" is supposed to be used when your running on your desktop/laptop. In Zephyr it is the "posix" board or the equivelent for Windows. If FreeRTOS has a posix mode why could they not implement metal_wait_usec???

That explains why the tests go into timeout. What I did not test is using `k_yield `with `k_can_yield `protection.
However, for the v2024.10 release, it is better to maintain the legacy behavior in all cases.

tnmysh (approved):
LGTM.

PR #324, opened 2025-02-11
io: fix buffer overflow in metal_sys_io_mem_map()

iuliana-prodan:
In metal_sys_io_mem_map() if the I/O region size
is a multiple of (1<<page_shift) will result in a
buffer overflow in the for loop.

Therefore, adjust loop termination condition to
prevent accessing an out-of-bounds page.

### DIFF
--- a/lib/system/generic/io.c
+++ b/lib/system/generic/io.c
@@ -22,7 +22,7 @@ void metal_sys_io_mem_map(struct metal_io_region *io)
 	if (psize) {
 		if (psize >> io->page_shift)
 			psize = (size_t)1 << io->page_shift;
-		for (p = 0; p <= (io->size >> io->page_shift); p++) {
+		for (p = 0; p <= ((io->size - 1) >> io->page_shift); p++) {
 			metal_machine_io_mem_map(va, io->physmap[p],
 						 psize, io->mem_flags);
 			va += psize;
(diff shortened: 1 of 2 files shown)


edmooring (changes requested):
Please fix the two compliance issues ("io->size -1" -> "io->size - 1"). Otherwise, this looks good to go.

lib/system/generic/io.c
```diff
@@ -22,7 +22,7 @@ void metal_sys_io_mem_map(struct metal_io_region *io)
 	if (psize) {
 		if (psize >> io->page_shift)
 			psize = (size_t)1 << io->page_shift;
-		for (p = 0; p <= (io->size >> io->page_shift); p++) {
+		for (p = 0; p <= ((io->size - 1) >> io->page_shift); p++) {
```
tnmysh, line 25: Can we fix this as following? 

p < (io->size >> io->page_shift)

@iuliana-prodan, @arnopo 
I don't have test to verify above logic, but if we reduce <= to only <, then we shouldn't need to do (io->size - 1) I think. I am okay with patch but was wondering if it could be optimized.

tnmysh: Thanks for the explanation.

tnmysh (approved):

edmooring (approved):
Looks good to go.

PR #332, opened 2025-05-07
Amd upstreaming commits may 7 2025

bentheredonethat:
AMD fork of Libmetla has been carrying commits for a while to ensure that the Libmetal AMP Demo for AMD-Xilinx SOC's boots on board with latest tools.
This series pushes patches that have been held for awhile.

Some of the patches are so that Libmetal Demo can be configured via Yocto build, others so demo works on now SOCs  and there are a few bug fixes.


bentheredonethat:
I have been told it is still ok to push changes to upstream libmetal demo area so here is PR for that.

bentheredonethat:
will need to clean up more before reviewing. sorry for churn here.

bentheredonethat:
still more warnings in early patches (very old patches at this point) to work through. will work to have those ready in next few days.

bentheredonethat:
for the first 4 commits (from Wendy years back) are you ok if I squash those and clean up as one commit?

arnopo:
Hello @bentheredonethat,
Could you provide more details in your PR description about the aim of this PR?
 You also need to add your sign-off in the commit, in addition to the author's.

bentheredonethat:
updated per your comment @arnopo

arnopo:
I would not be in favor of adding new demos in libmetal. I think this should be in the system reference. More than that, I would expect that we only integrate a few demos that provide examples to help developers implement Libmetal on their own platform.

 I propose to add the discussion to the agenda for next week's system reference meeting.

edmooring:
> I would not be in favor of adding new demos in libmetal. I think this should be in the system reference. More than that, I would expect that we only integrate a few demos that provide examples to help developers implement Libmetal on their own platform.
> 
> I propose to add the discussion to the agenda for next week's system reference meeting.

I am inclined to agree with you. Was there any resolution in the meeting?

nathalie-ckc:
Hi @edmooring , from the meeting:

* Arnaud: Thought were not going to extend application for specific machine (maintain legacy ok)
* Ben: Not adding new application.  It is a big PR.  Can split it up.  It's extending support & robustness.
* Arnaud: Looks like a lot of C files added (e.g. ipi-mb.c)
* action: Ben to split up the libmetal PR 332 to be addressed feature-per-feature & demo-per-demo

https://github.com/OpenAMP/openamp-system-reference/wiki/OpenAMP-System-Reference-Meeting-Notes-2025

PR #335, opened 2025-05-19
Add cross-platform tests for libmetal

lntaf:
To-be-expanded suite of tests for libmetal using system-agnostic interfaces instead of system specific tests.

There's more tests coming to expand the suite to other interfaces the library provides for completeness.
As much as possible tests isolate their libmetal usage to the filename of the interface they're testing.
When unavoidable, as is the case with init, an extra test is provided testing just that interface to help narrow down any issues.

Known issues:
mutex and shmem testing are pthread based, this will be changed to metal_run in a followup patch.
Interfaces: device, irq, threads, atomics are missing and will also be added later if there's interest in this.

As such this directory doesn't yet tamper with any pre-existing files and uses a simple Makefile for building, after libmetal is built.

### DIFF
--- a/test/system/cross-platform/src/alloc.c
+++ b/test/system/cross-platform/src/alloc.c
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2025, CARV ICS FORTH.
+ *
+ * SPDX-License-Identifier:    BSD-3-Clause
+ */
+
+/*
+ * @file		alloc.c
+ * @brief		Cross-platform tests for alloc interface
+ */
+
+int test_alloc(void)
+{
+	struct metal_init_params metal_param = METAL_INIT_DEFAULTS;
+	int *mem_ptr;
+
+	metal_init(&metal_param);
+	metal_set_log_handler(metal_default_log_handler);
+	metal_set_log_level(METAL_LOG_ERROR);
+
+	mem_ptr = metal_allocate_memory(16);
+	if (!mem_ptr) {
+		metal_log(METAL_LOG_ERROR, "metal_allocate_memory(16) returned null\n");
+		return 1;
+	}
+
+	metal_log(METAL_LOG_DEBUG,
+		  "Writing -2147483648, -1, 0, 2147483647 to allocated memory\n");
+
+	mem_ptr[0] = -2147483648;
+	mem_ptr[1] = -1;
+	mem_ptr[2] = 0;
+	mem_ptr[3] = 2147483647;
+
+	if (mem_ptr[0] != -2147483648 || mem_ptr[1] != -1 ||
+	    mem_ptr[2] != 0 || mem_ptr[3] != 2147483647) {
+
+		metal_log(METAL_LOG_ERROR,
+			  "Allocated memory has values %d, %d, %d, %d instead of -2147483648, -1 , 0, 2147483647\n",
+			  mem_ptr[0], mem_ptr[1], mem_ptr[2], mem_ptr[3]);
+		return 1;
+	}
+
+	metal_free_memory(mem_ptr);
+
+	metal_finish();
+	return 0;
+}
--- a/test/system/cross-platform/src/config.c
+++ b/test/system/cross-platform/src/config.c
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2025, CARV ICS FORTH.
+ *
+ * SPDX-License-Identifier:    BSD-3-Clause
+ */
+
+/*
+ * @file		config.c
+ * @brief		Cross-platform tests for config interface
+ */
+
+int test_config(void)
+{
+	struct metal_init_params metal_param = METAL_INIT_DEFAULTS;
+
+	metal_init(&metal_param);
+	metal_set_log_handler(metal_default_log_handler);
+	metal_set_log_level(METAL_LOG_INFO);
+
+	metal_log(METAL_LOG_INFO,
+		  "Metal version is: %d.%d.%d\n",
+		  METAL_VER_MAJOR, METAL_VER_MINOR, METAL_VER_PATCH);
+
+	metal_log(METAL_LOG_INFO,
+		  "In string form: %s\n",
+		  METAL_VER);
+
+	metal_log(METAL_LOG_INFO,
+		  "Metal system identifies as: %s\n",
+		  METAL_SYSTEM);
+
+	metal_log(METAL_LOG_INFO,
+		  "Metal processor identifies as: %s\n",
+		  METAL_PROCESSOR);
+
+	metal_log(METAL_LOG_INFO,
+		  "Metal machine identifies as: %s\n",
+		  METAL_MACHINE);
+
+	metal_finish();
+	return 0;
+}
--- a/test/system/cross-platform/src/log.c
+++ b/test/system/cross-platform/src/log.c
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2025, CARV ICS FORTH.
+ *
+ * SPDX-License-Identifier:    BSD-3-Clause
+ */
+
+/*
+ * @file		log.c
+ * @brief		Cross-platform tests for log interface
+ */
+
+int test_log(void)
+{
+	struct metal_init_params metal_param = METAL_INIT_DEFAULTS;
+
+	metal_init(&metal_param);
+
+	metal_set_log_handler(metal_default_log_handler);
+	metal_set_log_level(METAL_LOG_WARNING);
+
+	metal_log(METAL_LOG_DEBUG,
+		  "Metal log level is: %d\n",
+		  metal_get_log_level());
+
+	metal_log(METAL_LOG_EMERGENCY, "Emergency log example\n");
+	metal_log(METAL_LOG_ALERT, "Alert log example\n");
+	metal_log(METAL_LOG_CRITICAL, "Critical log example\n");
+	metal_log(METAL_LOG_ERROR, "Error log example\n");
+	metal_log(METAL_LOG_WARNING, "Warning log example\n");
+	metal_log(METAL_LOG_NOTICE, "Notice log example\n");
+	metal_log(METAL_LOG_INFO, "Info log example\n");
+	metal_log(METAL_LOG_DEBUG, "Debug log example\n");
+
+	metal_finish();
+	return 0;
+}
--- a/test/system/cross-platform/src/mutex.c
+++ b/test/system/cross-platform/src/mutex.c
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2025, CARV ICS FORTH.
+ *
+ * SPDX-License-Identifier:    BSD-3-Clause
+ */
+
+/*
+ * @file		mutex.c
+ * @brief		Cross-platform tests for mutex interface, pthread based
+ */
+
+#define ARRAY_SIZE 100
+
+int sum_m;
+metal_mutex_t mut;
+
+struct args_m  {
+	int *arr;
+	int len1;
+	int len2;
+};
+
+void *array_sum_m(void *a)
+{
+	struct args_m *arg = (struct args_m *)a;
+	int len1 = arg->len1;
+	int len2 = arg->len2;
+	int mid = (len2 - len1) / 2;
+
+	for (int i = 0; i <= mid; i++) {
+		metal_mutex_acquire(&mut);
+
+		if ((len2 - len1 + 1) % 2 && i == mid) {
+			sum_m += arg->arr[len1 + i];
+			metal_log(METAL_LOG_DEBUG,
+				  "Sum is now %d from adding %d\n",
+				  sum_m, arg->arr[len1 + i]);
+			metal_mutex_release(&mut);
+			break;
+		}
+		sum_m += arg->arr[len1 + i] + arg->arr[len2 - i];
+		metal_log(METAL_LOG_DEBUG,
+			  "Sum is now %d from adding %d,%d\n",
+			  sum_m, arg->arr[len1 + i], arg->arr[len2 - i]);
+		metal_mutex_release(&mut);
+		usleep(1);
+	}
+
+	return NULL;
+}
+
+int test_mutex(void)
+{
+	struct metal_init_params metal_param = METAL_INIT_DEFAULTS;
+	int array[ARRAY_SIZE];
+
+	for (int i = 0; i < ARRAY_SIZE; i++) {
+		array[i] = i + 1;
+	}
+
+	int length = sizeof(array) / sizeof(int);
+	struct args_m args1 = {array, length / 2, length - 1};
+	struct args_m args2 = {array, 0, length / 2 - 1};
+	pthread_t tid1, tid2;
+
+	metal_init(&metal_param);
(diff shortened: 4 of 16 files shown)


lntaf:
Hello,
Yes that is exactly my goal. We needed a testing suite that utilizes the exact same top-level calls, prints and error codes independent of a machine's architecture or OS so it's easier to debug what went wrong. These tests being system independent would also allow someone to implement a generic machine's calls for the APIs provided by libmetal without also having to provide tests for them under the test/system directory. I've so far used the suite on:

Ubuntu 22.04.3 LTS (Linux 5.15.153.1-microsoft-standard-WSL2 x86_64)
Linux 5.10.4-tag- (Sophgo patch kernel, aarch64)
Linux 5.10.4-tag- (Sophgo patch kernel, riscv64)

All of these systems compiled natively using the provided makefile which currently builds as shown:
`gcc -g -I <top libmetal dir>/install/usr/local/include <top libmetal dir>/test/system/src/test_all.c \
-o <top libmetal dir>/test/system/src/test_all.out -L <top libmetal dir>/install/usr/local/lib -lmetal`

test_all.h includes all the individual testfiles and a small struct with the name of each one (for printing purposes) as well as storing the return value of each.
test_all.c is just the main testing source file handling all the calls and printing.
We compile test_all.c from src/ -> test_all.out in out/ directory which compiles each individual source file from the header includes.

I'll work on getting you a hierarchy soon because I'm still working on a couple that may be closer to completion so they could be included. When I submit my followup edit with possibly a few more tests and changing the hierarchy should I add a CMake rule that falls under the -DWITH_TESTS flag?

Any other feedback is much appreciated!

test/system/cross-platform/src/alloc.c
```diff
@@ -0,0 +1,48 @@
...
+int test_alloc(void)
+{
+	struct metal_init_params metal_param = METAL_INIT_DEFAULTS;
+	int *mem_ptr;
+
+	metal_init(&metal_param);
+	metal_set_log_handler(metal_default_log_handler);
+	metal_set_log_level(METAL_LOG_ERROR);
+
+	mem_ptr = metal_allocate_memory(16);
```
edmooring, line 21: Nit. 16 => 4*sizeof(int)

test/system/cross-platform/src/mutex.c
```diff
@@ -0,0 +1,93 @@
...
+		       &args1);
+
+	pthread_create(&tid2,
+		       NULL,
+		       array_sum_m,
+		       &args2);
+
+	pthread_join(tid1, NULL);
+	pthread_join(tid2, NULL);
+
```
edmooring, line 84: It would be nice if this could be abstracted away from the pthread API so this test could run on FreeRTOS, NuttX, and Zephyr as well.

That could be a future enhancement.

test/system/cross-platform/src/sleep.c
```diff
@@ -0,0 +1,42 @@
...
+
+	ret = metal_sleep_usec(SLEEP_USECONDS);
+	if (ret)
+		metal_log(METAL_LOG_ERROR,
+			  "Function metal_sleep_usec returned %d\n",
+			  ret);
+
+	metal_log(METAL_LOG_INFO,
+		  "Time is now %d\n",
+		  time(0));
```
edmooring, line 39: Maybe use metal_get_timestamp() instead of time()?

test/system/cross-platform/src/mutex.c
lntaf: I assume that would mean using metal_run? 
I could modify the code to use metal_run from the metal-test.h header file but I can't see an implementation of metal_run for NuttX or Zephyr.
Do I use that API anyway and the NuttX/Zephyr implementation of it would come later? I have no personal experience with NuttX so I don't know how easy it would be for me to implement a similar interface.

arnopo: Yes using `metal_run` seems a good solution, could you use a metal_test prefix for the function to differenciate test from libmetal API?.  You can create a `metal_run` for the system you don't test, that prints a "not implemented/supported" message  and returns an error

test/system/cross-platform/src/alloc.c
```diff
@@ -0,0 +1,48 @@
...
+	if (!mem_ptr) {
+		metal_log(METAL_LOG_ERROR, "metal_allocate_memory(16) returned null\n");
+		return 1;
+	}
+
+	metal_log(METAL_LOG_DEBUG,
```
arnopo, line 27: Nitpicking:
you could use `metal_err` and similar helpers for the logs

test/system/cross-platform/src/config.c
```diff
@@ -0,0 +1,42 @@
...
+		  "Metal system identifies as: %s\n",
+		  METAL_SYSTEM);
+
+	metal_log(METAL_LOG_INFO,
+		  "Metal processor identifies as: %s\n",
+		  METAL_PROCESSOR);
+
+	metal_log(METAL_LOG_INFO,
+		  "Metal machine identifies as: %s\n",
+		  METAL_MACHINE);
```
arnopo, line 38: Still nitpicking:
Don't hesitate to use as much as posible the 100 char per line. Not mandatory but ithat increase the readablity.

test/system/cross-platform/src/log.c
```diff
@@ -0,0 +1,36 @@
...
+		  metal_get_log_level());
+
+	metal_log(METAL_LOG_EMERGENCY, "Emergency log example\n");
+	metal_log(METAL_LOG_ALERT, "Alert log example\n");
+	metal_log(METAL_LOG_CRITICAL, "Critical log example\n");
+	metal_log(METAL_LOG_ERROR, "Error log example\n");
+	metal_log(METAL_LOG_WARNING, "Warning log example\n");
+	metal_log(METAL_LOG_NOTICE, "Notice log example\n");
+	metal_log(METAL_LOG_INFO, "Info log example\n");
+	metal_log(METAL_LOG_DEBUG, "Debug log example\n");
```
arnopo, line 32: not neccessary to do it in this PR, but this test could be enhance by creating a loop to test the different log levels

test/system/cross-platform/src/mutex.c
```diff
@@ -0,0 +1,93 @@
...
+			metal_mutex_release(&mut);
+			break;
+		}
+		sum_m += arg->arr[len1 + i] + arg->arr[len2 - i];
+		metal_log(METAL_LOG_DEBUG,
+			  "Sum is now %d from adding %d,%d\n",
+			  sum_m, arg->arr[len1 + i], arg->arr[len2 - i]);
+		metal_mutex_release(&mut);
+		usleep(1);
+	}
```
arnopo, line 47: Could you add a  header file or a comment to explain the objective of the test. It’s not obvious at first glance.

test/system/cross-platform/src/sleep.c
```diff
@@ -0,0 +1,42 @@
...
+
+	metal_init(&metal_param);
+	metal_set_log_handler(metal_default_log_handler);
+	metal_set_log_level(METAL_LOG_ERROR);
+
+	metal_log(METAL_LOG_INFO,
+		  "Time is currently %d, sleeping for %u seconds\n",
+		  time(0), SLEEP_SECONDS);
+
+	tmp_time = time(0);
```
arnopo, line 29: `time` is not system agnostic. You can create a `metal_test` function to address it

test/system/cross-platform/src/config.c
lntaf: Also done! Also coming soon

arnopo:
@lntaf  It seems you have added a merge commit to your pull request but did not update the code.  You may need to push your updates again.  :smiley:

test/system/cross-platform/src/mutex.c
arnopo: As to the metal_test prefix I'm not quite sure I understand, would you like me to define **METAL_TEST__H** as is the case with metal-test.h or do you mean converting test_mutex (and all similar functions) to metal_test_mutex?

My point here is to rename for instance `metal_run`  to `metal_test_run` to differenciate libmetal API from the test API

test/system/cross-platform/src/mutex.c
lntaf: Rewrote it and added a description on the file for mutex and spinlock to make it clearer

test/CMakeLists.txt
```diff
@@ -67,5 +71,45 @@ else (WITH_ZEPHYR)
...
+#  if (WITH_STATIC_LIB)
+#    add_executable(test-all-static ${_src_all})
+#    target_link_libraries(test-all-static ${PROJECT_NAME}-static)
+#    target_compile_options(test-all-static PUBLIC ${_ec_flgs})
+#    add_dependencies(test-all-static ${PROJECT_NAME}-static)
+#    install(TARGETS test-all-static RUNTIME DESTINATION bin)
+#    if (WITH_TESTS_EXEC)
+#      add_test(test-all-static test-all-static)
+#    endif()
+#  endif()
```
arnopo, line 112: dead code to remove?

test/alloc.c
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright (c) 2025, CARV ICS FORTH.
+ *
+ * SPDX-License-Identifier:    BSD-3-Clause
+ */
+
+/*
+ * @file		alloc.c
+ * @brief		Cross-platform tests for alloc interface
+ */
+
```
arnopo, line 11: no `#include <metal/...>` ?
how it can build?

test/system/zephyr/threads.c
```diff
@@ -0,0 +1,38 @@
...
+#include <metal/sys.h>
+#include <metal/utilities.h>
+#include "metal-test.h"
+
+int metal_test_run(int threads, metal_thread_t child, void *arg)
+{
+	metal_unused(threads);
+	metal_unused(child);
+	metal_unused(arg);
+	metal_err("metal_test_run is currently not implemented for Zephyr\n");
```
arnopo, line 16: Seems that metal_err macro does not accept simple message without parameter... this is another topic. Not mandatory to fix it for this PR.
you probably can fix the warning using
`metal_err("%s is currently not implemented for Zephyr\n", __func__);`

test/test_all.h
```diff
@@ -0,0 +1,67 @@
...
+#include "cache.c"
+#include "config.c"
+#include "cpu.c"
+#include "init.c"
+#include "log.c"
+#include "mutex.c"
+#include "shmem.c"
+#include "sleep.c"
+#include "spinlock.c"
+#include "utilities.c"
```
arnopo, line 44: include C file should be replaced by extern declarations. Not clean to include c file even if it is for test.
```
extern int test_alloc(void);
...
```

test/test_all.h
```diff
@@ -0,0 +1,67 @@
...
+#include <metal/irq.h>
+#include <metal/log.h>
+#include <metal/mutex.h>
+#include <metal/shmem.h>
+#include <metal/sleep.h>
+#include <metal/spinlock.h>
+#include <metal/sys.h>
+#include <metal/time.h>
+#include <metal/utilities.h>
+#include "metal-test.h"
```
arnopo, line 31: Ok, I understand now why you don't include header files in the .c files. These includes should be placed in the test .c files. Moreover, it would be better if all test .c files compile independently.

then, if you declare struct test_entry tests[] directly in test/test_all.c, then this header file would become unnecessary.

lntaf: Understood, I'll change the includes and compilation behavior for the .c files and eliminate the header file test_all.h

test/sleep.c
```diff
@@ -0,0 +1,45 @@
...
+
+	tmp_time = metal_get_timestamp();
+
+	ret = metal_sleep_usec(SLEEP_USECONDS);
+
+	if (ret) {
+		metal_err("Function metal_sleep_usec returned %d\n", ret);
+		return 1;
+	}
+	tmp_time = tmp_time;
```
arnopo, line 41: tmp_time variable seems useless

test/system/generic/CMakeLists.txt
```diff
@@ -3,6 +3,9 @@ collect (PROJECT_LIB_TESTS alloc.c)
 collect (PROJECT_LIB_TESTS irq.c)
 collect (PROJECT_LIB_TESTS mutex.c)
 collect (PROJECT_LIB_TESTS atomic.c)
+collect (PROJECT_LIB_TESTS threads.c)
+
+collect (PROJECT_LIB_TEST_ALL threads.c)
```
arnopo, line 8: Seems not make sense to add threads test for generic as it is bare metal ( no OS).
could we remove it?

test/test-all.c
```diff
@@ -0,0 +1,67 @@
...
+	{"config.c",    test_config, 0},
+	{"cpu.c",       test_cpu, 0},
+	{"init.c",      test_init, 0},
+	{"log.c",       test_log, 0},
+	{"mutex.c",     test_mutex, 0},
+	{"shmem.c",     test_shmem, 0},
+	{"sleep.c",     test_sleep, 0},
+	{"spinlock.c",  test_spinlock, 0},
+	{"utilities.c", test_utilities, 0}
+};
```
arnopo, line 46: what about threads test?

test/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 collector_create (PROJECT_LIB_TESTS "${CMAKE_CURRENT_SOURCE_DIR}")
+collector_create (PROJECT_LIB_TEST_ALL "${CMAKE_CURRENT_SOURCE_DIR}")
```
arnopo, line 2: What about PROJECT_LIB_API_TEST project instead?

test/time.c
```diff
@@ -33,7 +33,7 @@ int test_time(void)
 		metal_err("Function metal_sleep_usec returned %d\n", ret);
 		return 1;
 	}
-
+	tmp_time = tmp_time;
```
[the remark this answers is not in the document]
lntaf: indeed it is, presumably a remnant from my own testing, apologies.

test/system/zephyr/threads.c
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2016, Xilinx Inc. and Contributors. All rights reserved.
```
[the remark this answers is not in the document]
lntaf: Yep, I left the xilinx copyright since to me it seemed more proper seeing how the entire threads.c file for zephyr is somewhat of a placeholder and was directed from this pull request's code review so it didn't seem entirely proper to take "credit" for a file with effectively no code :)
Nevertheless, I'll change it to CARV ICS FORTH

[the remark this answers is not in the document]
arnopo: In such case it is fair to  put both 

```
 * Copyright (c) 2025, CARV ICS FORTH
 * Copyright (c) 2016, Xilinx Inc. and Contributors. All rights reserved.
```

lntaf:
Apologies for the delay, I believe everything is in order now

arnopo:
> @arnopo , are you okay if we merge current patch as it is and do above enhancement after the release?
> 
> Thanks.
@tnmysh 
Yes it is too late to include this in this release as code is frozen. I plan to integrate it just after the release.

PR #338, opened 2025-06-05
add public header files for to metal-shared/static CMake targets

aelray:
For consumers who don't want to install libmetal it's more convenient to use the CMake targets metal-static or metal-shared. However, they currently don't bring the necessary include directories (In contrast to the Zephyr version where the include directories are already exported via `zephyr_include_directories(...)`)

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -89,6 +89,7 @@ else (WITH_ZEPHYR)
       VERSION           "${PROJECT_VERSION}"
       SOVERSION         "${PROJECT_VERSION_MAJOR}"
     )
+    target_include_directories (${_lib} PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/include)
   endif (WITH_SHARED_LIB)
 
   # Build a static library if so configured.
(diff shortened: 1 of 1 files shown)


aelray:
> Please fix the compliance check errors in your commit message. It needs a "Signed-off-by:" from your email address and the lines are too long.

Thanks for the review. I hope the commit message is fine now.

edmooring (changes requested):
The commit message still has a very visible typo "satic" should be "static".

aelray:
> The commit message still has a very visible typo "satic" should be "static".

Sorry for the typo. Fixed it but now I'm getting a compliance fail which is not so obvious to fix.

edmooring:
> > The commit message still has a very visible typo "satic" should be "static".
> 
> Sorry for the typo. Fixed it but now I'm getting a compliance fail which is not so obvious to fix.

This doesn't look like it's an issue with your commit. It looks like something broke in the test configuration. @arnopo can you take a look?

edmooring (approved):
Looks good to go.

arnopo (approved):

PR #339, opened 2025-06-25
lib: system: linux: Remove redundant memory validness check

foxhoundsk:
In the function prologue, "ldev" has already been assigned a malloc'd address, and a check has also been performed.  Remove this redundant pointer validness check during each iteration. The second check was also inherently misplaced because it was not enclosed in the malloc if-branch.


tnmysh (approved):
LGTM.

foxhoundsk:
> Only minor comment in commit message that, remove @ from variable name (ldev). Github is interpreting it as username.

Applied. Thanks!

arnopo (approved):

PR #341, opened 2025-06-30
lib: Add mailbox interface

bentheredonethat:
Add mailbox interface to init, deinit, send message and register callbacks.

Currently only implementation is for Zephyr.

Generic and FreeRTOS are just stubs.

CC @tnmysh


bentheredonethat:
updated per copilot comments CC @arnopo

lib/system/zephyr/mbox.c
```diff
@@ -0,0 +1,42 @@
...
+#include <metal/mbox.h>
+
+int __metal_mbox_send(metal_mbox_t *mbox, void *data, unsigned int len)
+{
+	metal_assert(mbox && mbox->m);
+	struct mbox_msg msg;
+	const int max_transfer_size_bytes = mbox_mtu_get_dt(mbox->m);
+
+	/* If data is NOT NULL, then pack message. */
+	if (data) {
```
arnopo, line 22: If I well remember  `data` can be not `NULL` for generic samples while the driver not support data so the `max_transfer_size_bytes != 0 ` should be tested first

lib/system/zephyr/mbox.h
```diff
@@ -0,0 +1,157 @@
...
+ * @brief	Zephyr mbox primitives for libmetal.
+ *
+ */
+
+#ifndef __METAL_MBOX_ZEPHYR__H__
+#define __METAL_MBOX_ZEPHYR__H__
+
+#include <metal/assert.h>
+#include <metal/sys.h>
+#include <zephyr/drivers/mbox.h>
```
arnopo, line 18: Is there any use case where Zephyr would call this metal API? How do you address the legacy IPM API?
I think we should forbid this API for Zephyr, as a Zephyr application should use the Zephyr mailbox or IPM API instead.

lib/system/generic/mbox.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * @brief	Initialize a libmetal mbox.
+ * @param[in]	mbox	mbox to initialize.
+ * @param[in]	init_data	Data used to initialize mbox.
+ *				In this case the driver instance is
+ *				already set up at Zephr init.
+ * @return	0 on success. Negative value on failure.
+ */
+extern int __metal_mbox_init(metal_mbox_t *mbox,
```
arnopo, line 39: Instead of using external define weak functions in a mbox.c that return error

arnopo (changes requested):
Regarding this PR, it seems to me that you need this API only to define a memory region in the Zephyr Device Tree associated with the mailbox.
If that is the case, I would suggest proposing support for an optional memory region in the Zephyr mailbox driver.
From my point of view, it does not make sense to add this API for Zephyr, which should only use Zephyr’s mailbox API.

Please notica also that the API implementation is missing for NuttX.

lib/system/generic/mbox.h
bentheredonethat: ok can do!

bentheredonethat:
> Regarding this PR, it seems to me that you need this API only to define a memory region in the Zephyr Device Tree associated with the mailbox. If that is the case, I would suggest proposing support for an optional memory region in the Zephyr mailbox driver. From my point of view, it does not make sense to add this API for Zephyr, which should only use Zephyr’s mailbox API.
> 
> Please notice also that the API implementation is missing for NuttX.

Hi @arnopo 

the APIs we wanted to introduce in generic function were send and receive using generic APIs

For send: As AMD supports buffered and bufferless interrupts (without payload) we want both in there and plan to upstream such support in the zephyr mailbox driver ecosystem.

For receive, mailbox zephyr API suports registering a callback. Again because of handling both buffered and bufferless both are included.

also here is references btw for zephyr send 
https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/drivers/mbox/remote/src/main.c#L62 
https://github.com/zephyrproject-rtos/zephyr/blob/main/samples/drivers/mbox_data/src/main.c#L58 

also i can add the nuttx implementation yes.

lib/system/zephyr/mbox.h
arnopo: I would like to avoid such implementation but having a more portable one (from my POV).
Having a generic API header file that declares API with one .c  file per system seems to me a better way. 
in other term I would like to avoid the use of ` #include <metal/system/@PROJECT_SYSTEM@/mbox.h>`

bentheredonethat:
Hi @arnopo 
> What I question here is whether you could or should manage the shared memory directly in your Zephyr mailbox driver instead of adding extra layers and introducing direct function calls between Zephyr and libmetal.

In the AMD-Xilinx case, the buffers are reserved for use by these specific interrupts. In our manuals they are called buffered inter-proccessor interrupts. The zephyr mailbox driver makes use of these specifically reserved buffers for the data passing. They are not arbitrarty ranges in memory.

 The AMD-Xilinx BSP also implements a mailbox driver for this.

Hopefully that answers your question.

arnopo:
> Hi @arnopo
> 
> > What I question here is whether you could or should manage the shared memory directly in your Zephyr mailbox driver instead of adding extra layers and introducing direct function calls between Zephyr and libmetal.
> 
> In the AMD-Xilinx case, the buffers are reserved for use by these specific interrupts. In our manuals they are called buffered inter-proccessor interrupts. The zephyr mailbox driver makes use of these specifically reserved buffers for the data passing. They are not arbitrarty ranges in memory.
> 
> The AMD-Xilinx BSP also implements a mailbox driver for this.
> 
> Hopefully that answers your question.

So I 'm lost.  I really don't understand why you need this API in zephyr. Calling external API in Zephyr applications look to me not a good idea if a Zephyr APi already exists.
 
what I would like to understand is:
- What is missing in Zephyr that requests a new API for AMD ?
- Why missing feature/checks can not be directly implemented in Zephyr?

From my point of view, this API should be allowed only for bare-metal and FreeRTOS.

bentheredonethat:
@arnopo  we would like to close this PR if you are OK with it.

Reason is that we want to first make it work in demo and then if needed we will port in in libmetal.

PR #343, opened 2025-10-08
Add QNX cross-compilation support 

dchordia:
Hi,
This contribution adds cross-compilation support for QNX operating system to libmetal.

### _Motivation_
QNX is an industry standard real-time operating system for embedded systems, particularly for vehicles, featuring safety focused features. It has also recently been made free for non-commercial use through the QNX Everywhere program, aiming for students, researchers and hobbyists to experiment with the OS.

Given that libmetal is a powerful abstraction layer across different OSes and baremetal as well as being a dependency to [open-amp](https://github.com/OpenAMP/open-amp), it would be great to enable libmetal to run natively on free version of QNX OS.

### _Description of changes_

Following API additions are made to support libmetal on QNX system
- allocate and free memory
- assert statements
- caching support
- condition variables
- physical devices like static shared memory
- i/o configurations
- interrupt requests
- logs
- mutex
- dynamic shared memory (under /dev/shmem) with precise page-wise controls sleep
- time
- system and misc utilities

These changes have been reviewed by QNX open-source team at [supporting PR](https://github.com/qnx-ports/libmetal/pull/1), for v2024.10.0, and additional supporting changes have been made to make it compatible with latest main.

### _How to build for QNX_
The build-files and instructions for cross-compiling libmetal for QNX are present at [qnx-ports](https://github.com/qnx-ports/build-files/tree/main/ports/libmetal). A free QNX8 license can be obtained [here](https://blackberry.qnx.com/en/products/qnx-everywhere)

By submitting this pull request, I confirm that my contribution is made under the terms of [BSD License](https://opensource.org/license/bsd-3-clause)

PS: Tagging @arnopo for review


lib/system/qnx/alloc.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2025, BlackBerry Limited. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/*
```
arnopo: To update in all source file in this PR
```suggestion
/**
```

lib/system/qnx/init.c
```diff
@@ -0,0 +1,79 @@
...
+	/* Initialize generic bus */
+	ret = metal_bus_register(&metal_generic_bus);
+	if (ret != 0) {
+		metal_log(METAL_LOG_DEBUG, "generic bus init failed - %s\n",
+			  strerror(-ret));
+	} else {
+		GENERIC_BUS_REGISTER = 1;
+	}
+
+	metal_unused(params);
```
arnopo: to move at the beginning of the function

arnopo (changes requested):
I started to review your PR and posted a few comments.

I’m not familiar with QNX, so I would like to understand if your porting could be simplified. 
It seems you use Linux as a reference system. This reference might not be the most appropriate since it more complex allowing the implementation of  the OpenAMP library in Linux userspace (with specific IRQ management, filesystem for shared memory, etc.).

If QNX is more similar to Zephyr, NuttX, or FreeRTOS, I suggest looking at those portings to simplify yours.

Additionally, please split your commit into two: one for the library and one for the tests.

elahav:
There are multiple issues with this port that make it incorrect.

1. QNX only ever works with virtual addresses. The MMU is always on. This affects, e.g., the cache code which seems to work on physical addresses.
2. Buffer allocation for DMA should be done by populating and then mapping shared-memory objects, not by calling `malloc()`. If you map as non-cached you don't need the cache manipulation. If you map cached then you need to flush caches using virtual addresses. The shared objects can then be queried for physical addresses if these are required in order to program hardware for DMA transfers.
3. I/O mapping needs to be done as device non-cached memory (nGnRnE on aarch64).
4. All code in QNX (outside of the micro-kernel) runs in non-privileged context, and therefore can't disable interrupts.

PR #346, opened 2025-10-21
[Fix for release] lib: system: re-enable the support of the template machine

arnopo:
The template machine is used in:
cmake/platforms/template-generic.cmake
cmake/platforms/template-freertos.cmake

This reverts a part of commit:
16493b179fb4 ("lib: system: remove xlnx BSP specific code")

### DIFF
--- a/lib/system/freertos/sys.h
+++ b/lib/system/freertos/sys.h
@@ -20,6 +20,7 @@
 #include <metal/errno.h>
 #include <metal/cpu.h>
 
+#include "./@PROJECT_MACHINE@/sys.h"
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/lib/system/freertos/template/CMakeLists.txt
+++ b/lib/system/freertos/template/CMakeLists.txt
@@ -1,0 +1,4 @@
+collect (PROJECT_LIB_HEADERS sys.h)
+
+collect (PROJECT_LIB_SOURCES sys.c)
+
--- a/lib/system/freertos/template/sys.c
+++ b/lib/system/freertos/template/sys.c
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2018, Linaro Inc. and Contributors. All rights reserved.
+ *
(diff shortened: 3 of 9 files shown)


bentheredonethat:
Hi @arnopo  thanks for this. one gotcha here is that freertos should be required in as part of the CMake Lists dependencies right?

example https://gitenterprise.xilinx.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/system/freertos/machine/zynqmp_r5/CMakeLists.txt#L2

arnopo:
> Hi @arnopo thanks for this. one gotcha here is that freertos should be required in as part of the CMake Lists dependencies right?
> 
> example https://gitenterprise.xilinx.com/OpenAMP/openamp-system-reference/blob/main/examples/legacy_apps/system/freertos/machine/zynqmp_r5/CMakeLists.txt#L2

Hi @bentheredonethat 
I have no access to your  https://gitenterprise.xilinx.com site ;-) 
I would propose to not force the dependency on Freertos, for the template.  Include path and library  could  also be provided during compilation time.

tnmysh:
@arnopo 

should we remove following template cmake files instead?

cmake/platforms/template-generic.cmake
cmake/platforms/template-freertos.cmake

Reason is: libmetal is expected to me machine independent.

We had introduced `template` machine, because `xlnx` machine was there.
Now since `xlnx` isn't there, I don't think `template` machine is needed.

arnopo:
> @arnopo
> 
> should we remove following template cmake files instead?
> 
> cmake/platforms/template-generic.cmake cmake/platforms/template-freertos.cmake
> 
> Reason is: libmetal is expected to me machine independent.
> 
> We had introduced `template` machine, because `xlnx` machine was there. Now since `xlnx` isn't there, I don't think `template` machine is needed.

I see the `template` more as an entry point to add your own machine. More than that at least ST uses it as a default machine since we have no need to add more configuration.

It is probably too late for this release, but we could transform the template `machine` into a `default` machine, with a deprecation mechanism for the template machine.

PR #350, opened 2025-11-19
Build machine-less libmetal

tnmysh:
It is not required to have MACHINE variable defined for the libmetal project. As Vendors can choose to implement machine specific interfaces outside of the libmetal and link it build time with the demos.

Hence make MACHINE variable optional in the cmake build system.


arnopo:
> The CI commands might need to be improved for this PR to pass:
> 
> cd /github/workspace/build-generic/lib && /usr/bin/arm-none-eabi-gcc -DDEFAULT_LOGGER_ON -DMETAL_INTERNAL -I/github/workspace/build-generic/lib/include -Wall -Werror -Wextra -MD -MT lib/CMakeFiles/metal-static.dir/dma.c.obj -MF CMakeFiles/metal-static.dir/dma.c.obj.d -o CMakeFiles/metal-static.dir/dma.c.obj -c /github/workspace/lib/dma.c In file included from /github/workspace/build-generic/lib/include/metal/sys.h:85, from /github/workspace/build-generic/lib/include/metal/io.h:22, from /github/workspace/build-generic/lib/include/metal/device.h:16, from /github/workspace/lib/dma.c:9: /github/workspace/build-generic/lib/include/metal/system/generic/sys.h:27:10: fatal error: ./template/sys.h: No such file or directory 27 | #include "./template/sys.h" | ^~~~~~~~~~~~~~~~~~
> 
> Since MACHINE is not mandatory anymore, may be we should explicitly pass it to build the library if needed.

Hi @tnmysh : sorry for the late reply.

From my perspective the CI is good we specify the `template` machine in `cmake/platforms/template-generic.cmake`.  The issue here is that the `build-generic/lib/include/metal/system/generic/template` folder is no more created.

Investigating the issue, it  seems that it is coming from a syntax error in the PR

```
-if (DEFINED ${PROJECT_MACHINE})
+if (DEFINED PROJECT_MACHINE)
-endif (DEFINED ${PROJECT_MACHINE})
+endif (DEFINED PROJECT_MACHINE) 
```

tnmysh:
> Something is not Crystal clear in AMD implementation What is the value of `PROJECT_SYSTEM` regarding https://github.com/OpenAMP/libmetal/blob/main/lib/sys.h#L85? I suppose that it is pointing to AMD sys.h defined in an AMD repo, right? i would be nice to add aditional information in documentation to explain such mechanism, that should become the recommended implementation.

HI @arnopo ,

sorry for delayed reply. 

https://github.com/OpenAMP/libmetal/blob/main/lib/sys.h#L85 here we are passing `PROJECT_SYSTEM` value, but we don't want to pass `PROJECT_MACHINE` value.

Here https://github.com/OpenAMP/libmetal/blob/1c3410be761e5048d00ada016c03335243b35d6c/lib/system/generic/sys.h#L27
we don't have any `PROJECT_MACHINE` so, no header will be included.

Instead, we will include metal/sys.h into our vendor specific extension module, and implement those interfaces directly there and link them during building demos.

arnopo:
Hi @tnmysh,

It is still not clear to me.

> https://github.com/OpenAMP/libmetal/blob/main/lib/sys.h#L85 here we are passing `PROJECT_SYSTEM` value, but we don't want to pass `PROJECT_MACHINE` value.

what is the value of `PROJECT_SYSTEM` in your environment?

> 
> Here
> 
> https://github.com/OpenAMP/libmetal/blob/1c3410be761e5048d00ada016c03335243b35d6c/lib/system/generic/sys.h#L27
> 
> 
> we don't have any `PROJECT_MACHINE` so, no header will be included.
> Instead, we will include metal/sys.h into our vendor specific extension module, and implement those interfaces directly there and link them during building demos.

You don't include the `metal/sys.h` generates from https://github.com/OpenAMP/libmetal/blob/main/lib/sys.h?

Perhaps a link to the AMD extension module you mention, would help me to understand how you split the libmetal

tnmysh:
> Hi @tnmysh,
> 
> It is still not clear to me.
> 
> > https://github.com/OpenAMP/libmetal/blob/main/lib/sys.h#L85 here we are passing `PROJECT_SYSTEM` value, but we don't want to pass `PROJECT_MACHINE` value.
> 
> what is the value of `PROJECT_SYSTEM` in your environment?

PRJOECT_SYSTEM can be one of the four: "generic", "freertos", "linux", or "zephyr".

> 
> > Here
> > https://github.com/OpenAMP/libmetal/blob/1c3410be761e5048d00ada016c03335243b35d6c/lib/system/generic/sys.h#L27
> > 
> > we don't have any `PROJECT_MACHINE` so, no header will be included.
> > Instead, we will include metal/sys.h into our vendor specific extension module, and implement those interfaces directly there and link them during building demos.
> 
> You don't include the `metal/sys.h` generates from https://github.com/OpenAMP/libmetal/blob/main/lib/sys.h?
> 

Yes, metal/sys.h is included in the extension as header.

> Perhaps a link to the AMD extension module you mention, would help me to understand how you split the libmetal

Here it is: https://gitenterprise.xilinx.com/embeddedsw/embeddedsw/blob/xlnx_rel_v2025.2/ThirdParty/sw_services/libmetal_xlnx_extension/src/lib/metal_xlnx_extension.c

Here is the README: https://gitenterprise.xilinx.com/embeddedsw/embeddedsw/blob/xlnx_rel_v2025.2/ThirdParty/sw_services/libmetal_xlnx_extension/src/lib/README.md

arnopo:
@tnmysh 

I had a look at your repo. Please tell me if my analysis is wrong.

- Regarding the [README](https://github.com/Xilinx/embeddedsw/tree/master/ThirdParty/sw_services/libmetal_xlnx_extension/src/lib/README.md)

  The MACHINE is set to "zynqmp_r5"  
`set (MACHINE "zynqmp_r5" CACHE STRING "")
`

- Regarding your [commit]( https://github.com/OpenAMP/libmetal/pull/350/changes/9797d1d0fef9334d53c01acb726f20356f69624e#diff-e8bc1b52c7618188b02120d369a4664f177556455e9a3c705d82b2cfd098d9c8R45)
 The `PROJECT_MACHINE`  is defined and set to  "zynqmp_r5"

  Including `libmetal/sys.h` will include `libmetal/zynqmp_r5/sys.h` refering to the code in https://github.com/OpenAMP/libmetal/blob/1c3410be761e5048d00ada016c03335243b35d6c/lib/system/generic/sys.h#L27

At this point I was expecting to  see libmetal/zynqmp_r5/sys.h in AMD extension

- Now regarding AMD extension seems that you hack [libmetal/generic/sys.h](https://github.com/Xilinx/embeddedsw/blob/master/ThirdParty/sw_services/libmetal/src/libmetal/lib/system/generic/sys.h#L25)

  It seems that you are replacing
  `#include "./@PROJECT_MACHINE@/sys.h" `
  by
  `#include <metal/system/generic/xlnx/sys.h>`

This indicates to me that you don't use the include files generated by CMake but rather those defined in the AMD extension.

If my analysis is correct, this highlights a limitation in libmetal that forces users to update libmetal itself.

I wonder if we should break the direct link between the system and machine include files.
By renaming the machine includes (for instance, from `sys.h` to `machine_sys.h `or `hal_sys.h`), and instead of defining the path in the #include, we should add it to the include path list. 

Is a fix is mandatory for do a fix for the v2025.10 release?

tnmysh:
> @tnmysh
> 
> I had a look at your repo. Please tell me if my analysis is wrong.
> 
> * Regarding the [README](https://github.com/Xilinx/embeddedsw/tree/master/ThirdParty/sw_services/libmetal_xlnx_extension/src/lib/README.md)
>   The MACHINE is set to "zynqmp_r5"
>   `set (MACHINE "zynqmp_r5" CACHE STRING "") `

This is different project than libmetal. So whatever MACHINE is defined here its not same as libmetal.

> * Regarding your [commit](https://github.com/OpenAMP/libmetal/pull/350/changes/9797d1d0fef9334d53c01acb726f20356f69624e#diff-e8bc1b52c7618188b02120d369a4664f177556455e9a3c705d82b2cfd098d9c8R45)
>   The `PROJECT_MACHINE`  is defined and set to  "zynqmp_r5"

No, libmetal has different toolchain.cmake, it is not same as above extension library.

>   Including `libmetal/sys.h` will include `libmetal/zynqmp_r5/sys.h` refering to the code in https://github.com/OpenAMP/libmetal/blob/1c3410be761e5048d00ada016c03335243b35d6c/lib/system/generic/sys.h#L27
> 
> At this point I was expecting to see libmetal/zynqmp_r5/sys.h in AMD extension
> 
> * Now regarding AMD extension seems that you hack [libmetal/generic/sys.h](https://github.com/Xilinx/embeddedsw/blob/master/ThirdParty/sw_services/libmetal/src/libmetal/lib/system/generic/sys.h#L25)
>   It seems that you are replacing
>   `#include "./@PROJECT_MACHINE@/sys.h" `
>   by
>   `#include <metal/system/generic/xlnx/sys.h>`
> 
> This indicates to me that you don't use the include files generated by CMake but rather those defined in the AMD extension.
> 

@arnopo 

Above is partially correct. Actually, I don't pass template-toolchain.cmake, instead I have defined my own toolchain file for `libmetal` where PROJECT_MACHINE is not defined at all. If you prefer I can include my toolchain file in this PR. When you don't define PROJECT_MACHINE and build libmetal with this patch, it simply ignores `@PROJECT_MACHINE/sys.h`. 

> If my analysis is correct, this highlights a limitation in libmetal that forces users to update libmetal itself.
> 

Correct. That is exactly why I introduced this patch, so extensions can be built out of libmetal.

> I wonder if we should break the direct link between the system and machine include files. By renaming the machine includes (for instance, from `sys.h` to `machine_sys.h `or `hal_sys.h`), and instead of defining the path in the #include, we should add it to the include path list.
> 

That is what this patch is doing. It removes requirement of @PROJECT_MACHINE/sys.h file.

> Is a fix is mandatory for do a fix for the v2025.10 release?

This fix is needed because we are forcing template-toolchain.cmake to build the libmetal. So yes it is needed for v2025.10. I think the last patch re-introduced template directory, and when I tested, I didn't have the last patch of v2025.10. That is why I missed it.

arnopo:
> > > Above is partially correct. Actually, I don't pass template-toolchain.cmake, instead I have defined my own toolchain file for `libmetal` where PROJECT_MACHINE is not defined at all. If you prefer I can include my toolchain file in this PR. When you don't define PROJECT_MACHINE and build libmetal with this patch, it simply ignores `@PROJECT_MACHINE/sys.h`.
> > 
> > 
> > from my test you don't ignore `@PROJECT_MACHINE/sys.h`. the `PROJECT_MACHINE` is just empty and in the generated [libmetal/lib/system/generic/sys.h](https://github.com/OpenAMP/libmetal/blob/1c3410be761e5048d00ada016c03335243b35d6c/lib/system/generic/sys.h#L27)
> > The result is `#include ".//sys.h" `
> 
> Correct, and that sys.h won't be included again because, it's guarded with #ifndef. So it's like as if that line is not taking effect.

Yes but the  `#include ".//sys.h" ` itself is not clean and the include of sys.h that is included by sys.h  is not clean and can introduce  confusion. 

> > 
> > For legacy support, we need to at least use deprecation mechanism for next release
> 
> Legacy support on AMD-xlnx platform ? Legacy support was introducing something xlnx specific, and not using standard libmetal API. With new extension, standard libmetal API use is possible. So, I am okay if we do not have legacy support on AMD-xlnx platform.

Not  only AMD platform.  The generic template machine  can be used as default machine for platforms that only needs metal for open-amp porting. This is the case for instance for ST platforms.

> 
> > > > I wonder if we should break the direct link between the system and machine include files. By renaming the machine includes (for instance, from `sys.h` to `machine_sys.h `or `hal_sys.h`), and instead of defining the path in the #include, we should add it to the include path list.
> > > 
> > > 
> > > That is what this patch is doing. It removes requirement of @PROJECT_MACHINE/sys.h file.
> > > > Is a fix is mandatory for do a fix for the v2025.10 release?
> > > 
> > > 
> > > This fix is needed because we are forcing template-toolchain.cmake to build the libmetal. So yes it is needed for v2025.10. I think the last patch re-introduced template directory, and when I tested, I didn't have the last patch of v2025.10. That is why I missed it.
> > 
> > 
> > Ok that means that we have to find a short term solution and probably rework it for next release
> > For this release, what about having in libmetal/lib/system/xxxx/sys.h
> 
> Do you mean, introducing xlnx/sys.h and xlnx/sys.c ? If we do that, then I will have to re-introduce xlnx directory under generic. As legacy support is not expected on AMD-xlnx platform, I prefer not to introduce "xlnx" direcotry again. We can have template directory, for legacy support.

I mean by `libmetal/lib/system/xxxx/sys.h`  `metal/system/@PROJECT_SYSTEM@/sys.h `  

> 
> > ```
> > #ifndef EXTERNAL_MACHINE
> > #include "./" PROJECT_MACHINE "/sys.h"
> > #endif
> > ```

apologogize  the code should be

```
 #ifndef EXTERNAL_MACHINE
 #include "./@PROJECT_MACHINE@/sys.h" 
 #endif

```
But writing this I realize that it is not clean, the generated code would be in AMD case
```
 #ifndef EXTERNAL_MACHINE
 #include ".//sys.h" 
 #endif

```
So forget this approach

What about following the approach done for the cache ?
- in `libmetal/lib/system/generic/sys.h` and `libmetal/lib/system/freertos/sys.h`
  -  remove line 
 ` #include "./@PROJECT_MACHINE@/sys.h" 
`
  - move fonctions declaration from  `lib/system/freertos/template/sys.h` and ` lib/system/generic/template/sys.h` to  `libmetal/lib/system/generic/sys.h` and `libmetal/lib/system/freertos/sys.h`
```
void sys_irq_enable(unsigned int vector);
void sys_irq_disable(unsigned int vector);
```
In such case we could remove  `lib/system/freertos/template/sys.h` and ` lib/system/generic/template/sys.h`

Could you try this with your AMD code?
I will test the template machine on my side.

> 
> >     
> >   
> > and in cmake
> 
> Which cmake? option.cmake ? or CMakeLists.txt in system directory?
> 
> Also, If above #include becomes no-op as explained above, then do we need extra definition?
> 
> > ```
> > if(NOT DEFINED PROJECT_MACHINE OR PROJECT_MACHINE STREQUAL "")
> >     add_definitions(-DEXTERNAL_MACHINE)
> > endif()
> > ```

tnmysh:
@arnopo pushed minor fix in your commit related to checkpatch.pl. Other than that this PR looks good.

xiaoxiang781216 (approved):

arnopo (approved):
same to #354 but for the main branch

arnopo:
I think we can ignor CI Issue for Zephyr build with Latest  version. It seems linked to zephyr infra that does not install python jsonschema module

PR #357, opened 2026-03-19
cmake: Do not propagate compilation options to the main project

arnopo:
When libmetal is included in the main project using add_subdirectory(libmetal), using set_property(GLOBAL ...) propagates properties to the main project. This can cause build errors if warnings are treated as errors.

Remove set_property(GLOBAL ...) calls from libmetal’s CMakeLists. Use only target_compile_options and target properties on libmetal’s own targets. The main project should determine its own compilation options.

fix commit d077e286a437 ("cmake: set PROJECT_EC_FLAGS to be GLOBAL property")

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -71,7 +71,7 @@ option (WITH_FUNC_LINE_LOG "Log with function name, line number prefix" OFF)
 
 option (WITH_DOC "Build with documentation" ON)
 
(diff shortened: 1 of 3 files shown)


tnmysh (approved):
LGTM.

tnmysh (approved):

tnmysh (approved):
LGTM.

edmooring (commented):
Looks good to go.

PR #360, opened 2026-04-20
lib: Avoid mentioning the current filename in compile-time warning

glneo:
Several of these were not correct in some spots. Use the automatic replacement to fill this in when installing the headers.

### DIFF
--- a/lib/system/freertos/alloc.h
+++ b/lib/system/freertos/alloc.h
@@ -10,7 +10,7 @@
  */
 
 #ifndef __METAL_ALLOC__H__
-#error "Include metal/alloc.h instead of metal/freertos/alloc.h"
+#error "Do not include this file directly, include <metal/alloc.h> instead"
 #endif
 
 #ifndef __METAL_FREERTOS_ALLOC__H__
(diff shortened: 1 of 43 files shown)


arnopo:
I would prefer to limit the use of cmake variable in source code when possible. 

Could you detail side effect you observe?

glneo:
What happens here for example is if I `#include <metal/system/freertos/alloc.h>`, the compiler errors out with: `Include metal/alloc.h instead of metal/freertos/alloc.h`. Which would be more helpful if I had included `metal/freertos/alloc.h` anywhere, I could grep that and fix it, but the included file is `metal/system/freertos/alloc.h`.

Maybe an easier solution is just to have the message state: "`Do not include this file directly, include <metal/alloc.h> instead`". Then folks could just see what file they included from the compiler error line and remove that inclusion.

edmooring (commented):
I like your second solution avoiding the CMake variables in the code better.

arnopo:
> Maybe an easier solution is just to have the message state: "`Do not include this file directly, include <metal/alloc.h> instead`". Then folks could just see what file they included from the compiler error line and remove that inclusion.

+1 for this solution

arnopo (approved):

arnopo:
@edmooring , @tnmysh : this commit would need a second review to be merged

tnmysh (approved):

PR #365, opened 2026-05-06
libmetal: lib: linux: improve Linux UIO-backed device open and test coverage

bentheredonethat:
This series improves the Linux UIO-backed device-open flow in libmetal and       
  adds test coverage for the new API and Linux-specific helper paths.              
                                                                                   
  The immediate motivation is to support Linux userspace applications that open 
  UIO-exposed devices through libmetal while keeping the existing bus/device       
  contract intact. In the current Linux implementation, the basic UIO path is   
  already present, but the backend is tightly coupled to bus probing, does not  
  cleanly separate resolved Linux device identities, does not unregister Linux  
  IRQ device state on close, and does not correctly retain the raw mapping         
  needed when a UIO map uses a non-zero offset.                                    
                                                                                   
  This series addresses those gaps in three steps:                                 
                                                                                   
    1. Add an explicit public helper,                                              
       metal_device_open_from_bus(), for bus/device based open while keeping       
       metal_device_open() as a compatibility wrapper.                             
                                                                                   
    2. Refactor the Linux UIO backend so it can populate a libmetal device         
       after resolving either a bus device name or a UIO class name, track the  
       resolved Linux identities more clearly, validate UIO map offsets, retain 
       raw mmap pointers for correct unmap, unregister Linux IRQ device state   
       on close, and tolerate Linux bus probe failure during metal_sys_init().  
                                                                                   
    3. Add cross-platform and Linux-specific tests for the new device-open         
       helper and the Linux UIO/IRQ bookkeeping helpers.                           
                                                                                   
  With these changes, downstream Linux host applications can continue to use       
  libmetal's device-open and IRQ registration model while relying on the           
  improved Linux UIO device handling in the library.

### DIFF
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -428,7 +428,8 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 	struct linux_bus *lbus = to_linux_bus(bus);
 	struct linux_device *ldev = NULL;
 	struct linux_driver *ldrv;
-	int error;
+	int error = -ENODEV;
+	int ret;
 
 	ldev = malloc(sizeof(*ldev));
 	if (!ldev)
@@ -448,9 +449,18 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 		ldev->device.bus = bus;
 
 		/* Try and open the device. */
-		error = ldrv->dev_open(lbus, ldev);
-		if (error) {
-			ldrv->dev_close(lbus, ldev);
+		ret = ldrv->dev_open(lbus, ldev);
+		if (ret) {
+			/*
+			 * Preserve the first useful errno while still allowing
+			 * clean backend misses to try the same device.
+			 */
+			if (ldrv->dev_close)
+				ldrv->dev_close(lbus, ldev);
+			if (error == -ENODEV)
+				error = ret;
+			if (ret != -ENODEV)
+				goto out;
 			continue;
 		}
 
@@ -461,9 +471,10 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 		return 0;
 	}
 
+out:
 	free(ldev);
 
-	return -ENODEV;
+	return error;
 }
 
 static void metal_linux_dev_close(struct metal_bus *bus,
@@ -668,4 +679,3 @@ int metal_linux_get_device_property(struct metal_device *device,
 	status = close(fd);
 	return status < 0 ? -errno : 0;
 }
-
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -462,12 +462,21 @@ static void metal_uio_dev_close(struct linux_bus *lbus,
 	(void)lbus;
 	unsigned int i;
 
+	if (ldev->fd >= 0) {
+		/*
+		 * Disable first so unregister only removes device bookkeeping;
+		 * IRQ handler teardown remains in the generic IRQ path.
+		 */
+		metal_irq_disable(ldev->fd);
+		metal_linux_irq_unregister_dev(ldev->fd);
+	}
 	for (i = 0; i < ldev->device.num_regions; i++) {
 		metal_unmap(ldev->region_map_raw[i],
 			    ldev->region_map_len[i]);
 		ldev->region_map_raw[i] = NULL;
 		ldev->region_map_len[i] = 0;
 	}
+	ldev->device.num_regions = 0;
 	if (ldev->override) {
 		sysfs_write_attribute(ldev->override, "", 1);
 		ldev->override = NULL;
@@ -478,6 +487,7 @@ static void metal_uio_dev_close(struct linux_bus *lbus,
 	}
 	if (ldev->fd >= 0) {
 		close(ldev->fd);
+		ldev->fd = -1;
 	}
 }
 
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -59,7 +59,13 @@ struct linux_device {
 	struct metal_device		device;
 	char				dev_name[PATH_MAX];
 	char				dev_path[PATH_MAX];
+	/*
+	 * UIO sysfs class directory, such as /sys/class/uio/uio0. UIO map
+	 * attributes are read relative to this path.
+	 */
 	char				cls_path[PATH_MAX];
+	char				uio_name[PATH_MAX];
+	char				uio_dev_name[PATH_MAX];
 	metal_phys_addr_t		region_phys[METAL_MAX_DEVICE_REGIONS];
 	void				*region_map_raw[METAL_MAX_DEVICE_REGIONS];
 	size_t				region_map_len[METAL_MAX_DEVICE_REGIONS];
@@ -122,6 +128,40 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
 
 	sysfs_close_attribute(attr);
 	return 0;
+}
+
+/**
+ * @internal
+ *
+ * @brief Read a string-valued UIO sysfs attribute.
+ *
+ * The value is explicitly terminated so callers can compare it as a C string.
+ *
+ * @param[in] path Path to the UIO sysfs attribute.
+ * @param[out] value Buffer that receives the attribute value.
+ * @param[in] len Size of value in bytes.
+ * @return 0 on success, or a negative error code on failure.
+ */
+static int metal_uio_read_str_attr(const char *path, char *value, size_t len)
+{
+	struct sysfs_attribute *attr;
+	int result = 0;
+
+	if (!value || !len)
+		return -EINVAL;
+
+	attr = sysfs_open_attribute(path);
+	if (!attr || sysfs_read_attribute(attr) != 0) {
+		result = -errno;
+		goto close_attr;
+	}
+
+	strncpy(value, attr->value, len - 1);
+	value[len - 1] = '\0';
+
+close_attr:
+	sysfs_close_attribute(attr);
+	return result;
 }
 
 /**
@@ -263,81 +303,30 @@ static int metal_uio_dev_bind(struct linux_device *ldev,
 	return 0;
 }
 
-static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
+/**
+ * @internal
+ *
+ * @brief Populate common UIO device state from resolved UIO paths.
+ *
+ * Both parent-bus opens and class-name opens share the same mmap, IRQ
+ * registration, DMA, and close-time cleanup rules.
+ *
+ * @param[in] lbus Linux bus used for diagnostics.
+ * @param[in,out] ldev Linux device with cls_path and dev_path already set.
+ * @return 0 on success, or a negative error code on failure.
+ */
+static int metal_uio_populate(struct linux_bus *lbus, struct linux_device *ldev)
 {
-	char *instance, path[SYSFS_PATH_MAX];
-	struct linux_driver *ldrv = ldev->ldrv;
 	unsigned long offset = 0, size = 0;
 	metal_phys_addr_t addr = 0, *phys;
 	struct metal_io_region *io;
 	struct metal_uio_map_info map_info;
 	size_t map_len, region_size;
-	struct dlist *dlist;
-	int result, i;
+	int result, i = 0;
 	unsigned int j;
 	void *raw, *virt;
 	int irq_info;
 
-
-	ldev->fd = -1;
-	ldev->device.irq_info = (void *)-1;
-
-	ldev->sdev = sysfs_open_device(lbus->bus_name, ldev->dev_name);
-	if (!ldev->sdev) {
-		metal_log(METAL_LOG_ERROR, "device %s:%s not found\n",
-			  lbus->bus_name, ldev->dev_name);
-		return -ENODEV;
-	}
-	metal_log(METAL_LOG_DEBUG, "opened sysfs device %s:%s\n",
-		  lbus->bus_name, ldev->dev_name);
-
-	result = metal_uio_dev_bind(ldev, ldrv);
-	if (result)
-		goto fail;
-
-	result = snprintf(path, sizeof(path), "%s/uio", ldev->sdev->path);
-	if (result < 0 || result >= (int)sizeof(path)) {
-		result = -EOVERFLOW;
-		goto fail;
-	}
-	dlist = sysfs_open_directory_list(path);
-	if (!dlist) {
-		metal_log(METAL_LOG_ERROR, "failed to scan class path %s\n",
-			  path);
-		result = -errno;
-		goto fail;
-	}
-
-	dlist_for_each_data(dlist, instance, char) {
-		result = snprintf(ldev->cls_path, sizeof(ldev->cls_path),
-				  "%s/%s", path, instance);
-		if (result < 0 || result >= (int)sizeof(ldev->cls_path)) {
-			result = -EOVERFLOW;
-			goto close_list;
-		}
-		result = snprintf(ldev->dev_path, sizeof(ldev->dev_path),
-				  "/dev/%s", instance);
-		if (result < 0 || result >= (int)sizeof(ldev->dev_path)) {
-			result = -EOVERFLOW;
-			goto close_list;
-		}
-		break;
-	}
-	result = 0;
-
-close_list:
-	sysfs_close_list(dlist);
-	if (result)
-		goto fail;
-
-	if (sysfs_path_is_dir(ldev->cls_path) != 0) {
-		metal_log(METAL_LOG_ERROR, "invalid device class path %s\n",
-			  ldev->cls_path);
-		result = -ENODEV;
-		goto fail;
-	}
-
-	i = 0;
 	do {
 		if (!access(ldev->dev_path, F_OK))
 			break;
@@ -347,14 +336,13 @@ close_list:
 	if (i >= 1000) {
 		metal_log(METAL_LOG_ERROR, "failed to open file %s, timeout.\n",
 			  ldev->dev_path);
-		result = -ENODEV;
-		goto fail;
+		return -ENODEV;
 	}
 	result = metal_open(ldev->dev_path, 0);
 	if (result < 0) {
-		metal_log(METAL_LOG_ERROR, "failed to open device %s\n",
+		metal_log(METAL_LOG_ERROR, "failed to open device %s: %s\n",
 			  ldev->dev_path, strerror(-result));
-		goto fail;
+		return result;
 	}
 	ldev->fd = result;
 
@@ -440,19 +428,136 @@ fail:
 	ldev->device.num_regions = 0;
 	ldev->device.irq_num = 0;
 	ldev->device.irq_info = (void *)-1;
-	if (ldev->override) {
-		sysfs_write_attribute(ldev->override, "", 1);
-		ldev->override = NULL;
-	}
-	if (ldev->sdev) {
-		sysfs_close_device(ldev->sdev);
-		ldev->sdev = NULL;
-	}
 	if (ldev->fd >= 0) {
 		close(ldev->fd);
 		ldev->fd = -1;
 	}
 
+	return result;
+}
+
+static void metal_uio_dev_close(struct linux_bus *lbus,
+				struct linux_device *ldev);
+
+/**
+ * @internal
+ *
+ * @brief Open a parent-bus device that exposes a UIO child.
+ *
+ * This path binds the parent device to a UIO driver before using the common
+ * UIO populate logic.
+ *
+ * @param[in] lbus Linux bus containing the parent device.
+ * @param[in,out] ldev Linux device to open.
+ * @return 0 on success, or a negative error code on failure.
+ */
+static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
+{
+	char *instance, path[SYSFS_PATH_MAX];
+	struct linux_driver *ldrv = ldev->ldrv;
+	struct dlist *dlist;
+	int result;
+
+	ldev->fd = -1;
+	ldev->device.irq_info = (void *)-1;
+
+	ldev->sdev = sysfs_open_device(lbus->bus_name, ldev->dev_name);
+	if (!ldev->sdev) {
+		metal_log(METAL_LOG_ERROR, "device %s:%s not found\n",
+			  lbus->bus_name, ldev->dev_name);
+		return -ENODEV;
+	}
+	metal_log(METAL_LOG_DEBUG, "opened sysfs device %s:%s\n",
+		  lbus->bus_name, ldev->dev_name);
+	/*
+	 * Error paths after this point clean up locally. The common open loop
+	 * may call dev_close() again, so close must tolerate partial cleanup.
+	 */
+
+	/*
+	 * Parent-bus opens still need the requested platform or PCI device
+	 * bound to the selected UIO driver before a /dev/uioX node can exist.
+	 */
+	result = metal_uio_dev_bind(ldev, ldrv);
+	if (result)
+		goto fail;
+
+	/*
+	 * A bound parent device exposes one UIO child below its sysfs device
+	 * directory. Use that child name to derive both sysfs and /dev paths.
+	 */
+	result = snprintf(path, sizeof(path), "%s/uio", ldev->sdev->path);
+	if (result < 0 || result >= (int)sizeof(path)) {
+		result = -EOVERFLOW;
+		goto fail;
+	}
+	dlist = sysfs_open_directory_list(path);
+	if (!dlist) {
+		metal_log(METAL_LOG_ERROR, "failed to scan class path %s\n",
+			  path);
+		result = -errno;
+		goto fail;
+	}
+
+	dlist_for_each_data(dlist, instance, char) {
+		/*
+		 * The first UIO child is the device node this parent-bus open
+		 * will use for mmap, IRQ, and DMA operations.
+		 */
+		result = snprintf(ldev->cls_path, sizeof(ldev->cls_path),
+				  "%s/%s", path, instance);
+		if (result < 0 || result >= (int)sizeof(ldev->cls_path)) {
+			result = -EOVERFLOW;
+			goto close_list;
+		}
+		result = snprintf(ldev->dev_path, sizeof(ldev->dev_path),
+				  "/dev/%s", instance);
+		if (result < 0 || result >= (int)sizeof(ldev->dev_path)) {
+			result = -EOVERFLOW;
+			goto close_list;
+		}
+		result = snprintf(path, sizeof(path), "%s/name", ldev->cls_path);
+		if (result < 0 || result >= (int)sizeof(path)) {
+			result = -EOVERFLOW;
+			goto close_list;
+		}
+		ldev->uio_name[0] = '\0';
+		metal_uio_read_str_attr(path, ldev->uio_name,
+					sizeof(ldev->uio_name));
+		result = snprintf(ldev->uio_dev_name,
+				  sizeof(ldev->uio_dev_name), "%s", instance);
+		if (result < 0 || result >= (int)sizeof(ldev->uio_dev_name)) {
+			result = -EOVERFLOW;
+			goto close_list;
+		}
+		break;
+	}
+
+	sysfs_close_list(dlist);
+	result = 0;
+
+	/* Refuse to continue if the selected UIO class path disappeared. */
+	if (sysfs_path_is_dir(ldev->cls_path) != 0) {
+		metal_log(METAL_LOG_ERROR, "invalid device class path %s\n",
+			  ldev->cls_path);
+		result = -ENODEV;
+		goto fail;
+	}
+
+	/*
+	 * Once cls_path and dev_path are resolved, the rest of the open flow is
+	 * shared with the synthetic UIO class-name path.
+	 */
+	result = metal_uio_populate(lbus, ldev);
+	if (result)
+		goto fail;
+
+	return 0;
+
+close_list:
+	sysfs_close_list(dlist);
+fail:
+	metal_uio_dev_close(lbus, ldev);
 	return result;
 }
 
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -111,6 +111,22 @@ static struct linux_device *to_linux_device(struct metal_device *device)
 /**
  * @internal
  *
+ * @brief Check whether a Linux bus is the synthetic UIO bus.
+ *
+ * The synthetic UIO bus opens devices directly from /sys/class/uio instead of
+ * through a sysfs bus directory and parent-bus drivers.
+ *
+ * @param[in] lbus Linux bus to check.
+ * @return true if the bus is the synthetic UIO bus, otherwise false.
+ */
+static bool metal_linux_is_uio_bus(const struct linux_bus *lbus)
+{
+	return strcmp(lbus->bus_name, "uio") == 0;
+}
+
+/**
+ * @internal
+ *
  * @brief Read the first text line from a sysfs file.
  *
  * The trailing newline is stripped so callers can compare sysfs text values as
@@ -553,8 +569,7 @@ static int metal_uio_populate(struct linux_bus *lbus, struct linux_device *ldev)
 	irq_info = 1;
 	if (write(ldev->fd, &irq_info, sizeof(irq_info)) <= 0) {
 		metal_log(METAL_LOG_INFO,
-			  "%s: No IRQ for device %s.\n",
-			  __func__, ldev->dev_name);
+			  "No IRQ for device %s.\n", ldev->dev_name);
 		ldev->device.irq_num =  0;
 		ldev->device.irq_info = (void *)-1;
 	} else {
@@ -850,6 +865,22 @@ static void metal_uio_dev_dma_unmap(struct linux_bus *lbus,
 
 static struct linux_bus linux_bus[] = {
 	{
+		.bus_name	= "uio",
+		.drivers = {
+			{
+				.drv_name  = "uio",
+				.mod_name  = "uio",
+				.cls_name  = "uio",
+				.dev_open  = metal_uio_class_dev_open,
+				.dev_close = metal_uio_dev_close,
+				.dev_irq_ack  = metal_uio_dev_irq_ack,
+				.dev_dma_map = metal_uio_dev_dma_map,
+				.dev_dma_unmap = metal_uio_dev_dma_unmap,
+			},
+			{ 0 /* sentinel */ }
+		}
+	},
+	{
 		.bus_name	= "platform",
 		.drivers = {
 			{
@@ -924,12 +955,18 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 	for_each_linux_driver(lbus, ldrv) {
 
 		/* Check if we have a viable driver. */
-		if (!ldrv->sdrv || !ldrv->dev_open)
+		if (!ldrv->dev_open ||
+		    (!metal_linux_is_uio_bus(lbus) && !ldrv->sdrv))
 			continue;
 
 		/* Reset device data. */
 		memset(ldev, 0, sizeof(*ldev));
-		strncpy(ldev->dev_name, dev_name, sizeof(ldev->dev_name) - 1);
+		error = snprintf(ldev->dev_name, sizeof(ldev->dev_name),
+				 "%s", dev_name);
+		if (error < 0 || error >= (int)sizeof(ldev->dev_name)) {
+			error = -EOVERFLOW;
+			goto out;
+		}
 		ldev->fd = -1;
 		ldev->ldrv = ldrv;
 		ldev->device.bus = bus;
@@ -974,6 +1011,16 @@ static void metal_linux_dev_close(struct metal_bus *bus,
 	free(ldev);
 }
 
+/**
+ * @internal
+ *
+ * @brief Close a Linux bus and any probed sysfs driver handles.
+ *
+ * The synthetic UIO bus has no sysfs bus handle, so the bus handle is closed
+ * only when one was opened during probing.
+ *
+ * @param[in] bus Metal bus to close.
+ */
 static void metal_linux_bus_close(struct metal_bus *bus)
 {
 	struct linux_bus *lbus = to_linux_bus(bus);
@@ -985,7 +1032,8 @@ static void metal_linux_bus_close(struct metal_bus *bus)
 		ldrv->sdrv = NULL;
 	}
 
-	sysfs_close_bus(lbus->sbus);
+	if (lbus->sbus)
+		sysfs_close_bus(lbus->sbus);
 	lbus->sbus = NULL;
 }
 
@@ -1084,10 +1132,27 @@ static int metal_linux_probe_driver(struct linux_bus *lbus,
 	return ldrv->sdrv ? 0 : -ENODEV;
 }
 
+/**
+ * @internal
+ *
+ * @brief Probe and register a Linux bus.
+ *
+ * The synthetic UIO bus is registered only when the /sys/class/uio class
+ * exists and does not use normal sysfs bus or driver probing.
+ *
+ * @param[in,out] lbus Linux bus to probe.
+ * @return 0 on success, or a negative error code on failure.
+ */
 static int metal_linux_probe_bus(struct linux_bus *lbus)
 {
 	struct linux_driver *ldrv;
 	int ret, error = -ENODEV;
+
+	if (metal_linux_is_uio_bus(lbus)) {
+		if (sysfs_path_is_dir(METAL_UIO_CLASS_PATH) != 0)
+			return -ENODEV;
+		return metal_linux_register_bus(lbus);
+	}
 
 	lbus->sbus = sysfs_open_bus(lbus->bus_name);
 	if (!lbus->sbus)
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -15,10 +15,13 @@
 #include <metal/irq.h>
 
 #include <stdint.h>
+#include <stdbool.h>
+#include <dirent.h>
 
 #include "irq.h"
 
 #define MAX_DRIVERS	64
+#define METAL_UIO_CLASS_PATH "/sys/class/uio"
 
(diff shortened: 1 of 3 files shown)


lib/system/linux/device.c
```diff
@@ -14,9 +14,13 @@
 #include <metal/utilities.h>
 #include <metal/irq.h>
 
+#include <stdbool.h>
```
arnopo, line 18: This commit is quite large and difficult to understand if you are not familiar with the AMD implementation based on UIO, which is my case.
- Could you split it into smaller, more readable commits?
- Does it make sense to keep UIO support  in the Linux subsystem, or is it specific to AMD?

lib/system/linux/device.c
```diff
@@ -14,9 +14,13 @@
...
+#include <dirent.h>
+
 #include "irq.h"
 
 #define MAX_DRIVERS	64
+#define METAL_UIO_CLASS_PATH "/sys/class/uio"
```
arnopo, line 24: Does UIO support is required  on the target in order to use OpenAMP in Linux user space?

lib/system/linux/irq.h
```diff
@@ -28,6 +28,9 @@
  * @param[in]	irq interrupt id
  */
 void metal_linux_irq_register_dev(struct metal_device *dev, int irq);
+void metal_linux_irq_unregister_dev(int irq);
+struct metal_device *metal_linux_irq_get_dev(int irq);
+int metal_linux_irq_is_enabled(int irq);
```
arnopo, line 57: APIs should have a doxygen documentation header

lib/system/linux/sys.h
```diff
@@ -104,18 +104,9 @@ int metal_map(int fd, off_t offset, size_t size, int expand,
...
- * @param[out]	output output buffer to store read data
- * @param[in]	len number of bytes to be read
- * @return	0 on success, or -errno on error.
- */
-int metal_linux_get_device_property(struct metal_device *device,
-				    const char *property_name,
-				    void *output, int len);
+int metal_linux_uio_validate_offset(const char *dev_name,
+				    unsigned int index,
+				    unsigned long offset);
```
arnopo: doxygen documentation header

lib/device.c
```diff
@@ -56,8 +56,8 @@ int metal_bus_find(const char *name, struct metal_bus **result)
 	return -ENOENT;
 }
 
-int metal_device_open(const char *bus_name, const char *dev_name,
-		      struct metal_device **device)
+int metal_device_open_from_bus(const char *bus_name, const char *dev_name,
```
arnopo: I do not understand the need to create a new API. It looks to me like this introduces confusion by duplicating APIs instead of making things more obvious.

Having a single system-agnostic API makes more sense to me.

arnopo (changes requested):

bentheredonethat:
Hi @arnopo 

  Thanks for the review. I agree the Linux change is too large as posted, and I can split it into smaller commits before respinning.

  The main problem I am trying to solve is narrower than the current series makes it look. Linux userspace libmetal already opens platform/PCI devices through the existing Linux UIO-backed path.
  Today applications have to pass the Linux bus device name, for example "ff360000.ipi". That name comes from the platform device/unit-address naming and is not a stable logical name across SoCs or
  even across different instances.

  For demos and portable userspace applications, we would like to use a logical name such as "demo-ipi". On systems using UIO, that logical name can be exposed by the kernel as the UIO class name,
  for example:

      /sys/class/uio/uioX/name = demo-ipi

  The Linux libmetal backend can then resolve that UIO class name back to the real bus device:

      demo-ipi
        -> /sys/class/uio/uioX/name
        -> /sys/class/uio/uioX/device
        -> /sys/bus/platform/devices/ff360000.ipi

  and continue through the existing Linux bus open/bind/map/IRQ/close flow.

  So the intended direction is not to make UIO mandatory for OpenAMP generally, and not to introduce a new Linux device model. It is only to let the existing Linux UIO backend accept the UIO class
  name as an alias for the native bus device name.

  I also agree with your comment about the new public API. The UIO-name use case does not require adding `metal_device_open_from_bus()`. I can drop that patch and keep the existing system-agnostic
  API:

      metal_device_open("platform", "demo-ipi", &dev);

  The Linux backend would interpret the device argument as either:
  1. the native bus device name, preserving existing behavior, or
  2. a UIO class-name alias, resolved internally to the native platform/PCI device.

  Internally I plan to keep both identities clear, e.g. requested name / resolved bus device name / UIO name. For minimum behavior change, the returned `device->name` can remain the resolved native
  bus device name; the UIO name is just an open-time alias.

  For the respin, does the following direction sound acceptable?

  1. Drop the new `metal_device_open_from_bus()` public API.
  2. Split the Linux changes into smaller commits, likely:
     - UIO mmap offset validation and correct raw-pointer unmap tracking.
     - IRQ bookkeeping cleanup on device close.
     - UIO class-name alias resolution for the existing Linux UIO backend.
     - Tests and documentation updates.
  3. Include Doxygen comments in the same commits that introduce new internal Linux helpers.
  4. Clarify in the commit messages that UIO-name lookup is optional alias resolution for the existing Linux backend, not a requirement for OpenAMP Linux userspace in general.

  If this direction makes sense, I will rework the series around that.

arnopo:
> Hi @arnopo
> 
> Thanks for the review. I agree the Linux change is too large as posted, and I can split it into smaller commits before respinning.
> 
> The main problem I am trying to solve is narrower than the current series makes it look. Linux userspace libmetal already opens platform/PCI devices through the existing Linux UIO-backed path. Today applications have to pass the Linux bus device name, for example "ff360000.ipi". That name comes from the platform device/unit-address naming and is not a stable logical name across SoCs or even across different instances.
> 
> For demos and portable userspace applications, we would like to use a logical name such as "demo-ipi". On systems using UIO, that logical name can be exposed by the kernel as the UIO class name, for example:
> 
> ```
>   /sys/class/uio/uioX/name = demo-ipi
> ```
> 
> The Linux libmetal backend can then resolve that UIO class name back to the real bus device:
> 
> ```
>   demo-ipi
>     -> /sys/class/uio/uioX/name
>     -> /sys/class/uio/uioX/device
>     -> /sys/bus/platform/devices/ff360000.ipi
> ```
> 
> and continue through the existing Linux bus open/bind/map/IRQ/close flow.

What about using symbolic link for that, as proposed by @tnmysh  in https://github.com/OpenAMP/openamp-system-reference/pull/101.  
That would avoid resolution by `/sys/class/uio/uioX/name`  if a `/sys/class/uio/<name>/device` symbolic is created with a udev rule.
Would it work in your case?

> 
> So the intended direction is not to make UIO mandatory for OpenAMP generally, and not to introduce a new Linux device model. It is only to let the existing Linux UIO backend accept the UIO class name as an alias for the native bus device name.
> 
> I also agree with your comment about the new public API. The UIO-name use case does not require adding `metal_device_open_from_bus()`. I can drop that patch and keep the existing system-agnostic API:
> 
> ```
>   metal_device_open("platform", "demo-ipi", &dev);
> ```
> 
> The Linux backend would interpret the device argument as either:
> 
> 1. the native bus device name, preserving existing behavior, or
> 2. a UIO class-name alias, resolved internally to the native platform/PCI device.
> 
> Internally I plan to keep both identities clear, e.g. requested name / resolved bus device name / UIO name. For minimum behavior change, the returned `device->name` can remain the resolved native bus device name; the UIO name is just an open-time alias.
> 
> For the respin, does the following direction sound acceptable?
> 
> 1. Drop the new `metal_device_open_from_bus()` public API.
> 2. Split the Linux changes into smaller commits, likely:
>    
>    * UIO mmap offset validation and correct raw-pointer unmap tracking.
>    * IRQ bookkeeping cleanup on device close.
>    * UIO class-name alias resolution for the existing Linux UIO backend.
>    * Tests and documentation updates.
> 3. Include Doxygen comments in the same commits that introduce new internal Linux helpers.
> 4. Clarify in the commit messages that UIO-name lookup is optional alias resolution for the existing Linux backend, not a requirement for OpenAMP Linux userspace in general.

Sound good. 

Thanks 
arnaud
> 
> If this direction makes sense, I will rework the series around that.

bentheredonethat:
Thanks @arnopo, this is a good point.

Using a udev-created symlink such as:

`    /sys/class/uio/<logical-name>/device
`
would work in our use case and is a nice optimization when present, since it avoids scanning `uioX/name`.

For upstream libmetal, I would prefer to treat this as optional platform integration rather than a hard dependency, because not all deployments guarantee custom udev rules. So my plan is:

  1) keep the generic fallback that resolves via `/sys/class/uio/uioX/name`,
  2) optionally try the symlink path first when it exists.

That keeps behavior portable out of the box while allowing integrators to use the symlink approach for faster/cleaner lookup.

If you agree, I will document this in the commit message as “optional udev optimization, generic fallback preserved”.

arnopo:
> Thanks @arnopo, this is a good point.
> 
> Using a udev-created symlink such as:
> 
> ` /sys/class/uio/<logical-name>/device` would work in our use case and is a nice optimization when present, since it avoids scanning `uioX/name`.
> 
> For upstream libmetal, I would prefer to treat this as optional platform integration rather than a hard dependency, because not all deployments guarantee custom udev rules. So my plan is:
> 
> 1. keep the generic fallback that resolves via `/sys/class/uio/uioX/name`,
> 2. optionally try the symlink path first when it exists.
> 
> That keeps behavior portable out of the box while allowing integrators to use the symlink approach for faster/cleaner lookup.
> 
> If you agree, I will document this in the commit message as “optional udev optimization, generic fallback preserved”.

I would prefer that we handle this in the same way we manage /dev/rpmsgX or /sys/class/remoteproc/remoteprocX devices, rather than adding it to libmetal.

I propose adding this PR to the agenda for the next OpenAMP meeting so we can discuss it further.

bentheredonethat:
Hi @arnopo  @wmamills  @tnmysh i have some updates here:

  Update: I pushed the revised libmetal changes to this PR.

The current branch now fleshes out the Linux `uio` bus support while preserving the pre-existing platform/device open flow. Existing users that open devices through the platform bus continue to use the same bind-and-open path. The new `uio` bus path is additive and lets userspace open already exposed UIO devices by their `/sys/class/uio/uioX/name` value, which gives applications a stable logical lookup path without requiring generated platform device names.

I also kept the fallback behavior in place: native platform-bus open remains the primary path for existing users, and the UIO class-name path is only used when callers explicitly request the `uio` bus. The shared UIO populate logic is reused so both paths get the same mmap offset handling, IRQ setup, DMA-map behavior, and cleanup.

I investigated the symlink-based lookup option as requested. I do not think symlinks are needed for this PR. The kernel already exposes the stable lookup key we need through `/sys/class/uio/uioX/name`, and resolving that directly avoids adding another filesystem convention that would need to be created, documented, kept in sync with UIO enumeration, and handled across distros/init systems/containers. Using the existing UIO class metadata keeps the implementation self-contained in libmetal and avoids requiring deployment-side symlink management.

So the PR now takes this approach:

  - Add explicit `uio` bus support for opening devices by UIO class name.
  - Preserve the existing platform bus behavior and fallback path.
  - Reuse the same UIO populate/mapping/IRQ cleanup logic for both paths.
  - Avoid symlink lookup because `/sys/class/uio/uioX/name` is sufficient and already available.

lib/system/linux/device.c
```diff
@@ -100,6 +102,72 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
...
+int metal_linux_uio_validate_offset(const char *dev_name,
+				    unsigned int index,
+				    unsigned long offset)
+{
+	const unsigned long page_size = (unsigned long)getpagesize();
+
+	if (offset >= page_size) {
+		metal_log(METAL_LOG_ERROR,
+			  "%s: device %s map%u has invalid UIO offset 0x%lx (page size 0x%lx)\n",
+			  __func__, dev_name ? dev_name : "<unknown>", index,
```
arnopo: remove `__func__ ` in your print `ML_FUNC_LINE` is used to enable such traces

to fix also in the rest of the PR

lib/system/linux/device.c
```diff
@@ -100,6 +102,72 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
...
+}
+
+int metal_linux_uio_map_info(const char *dev_name,
+			     unsigned int index,
+			     metal_phys_addr_t map_addr,
+			     unsigned long map_size,
+			     unsigned long offset,
+			     metal_phys_addr_t *phys,
+			     size_t *map_len,
+			     size_t *region_size)
```
arnopo: Define structures to avoid passing too many parameters.

lib/system/linux/sys.h
```diff
@@ -104,6 +104,47 @@ int metal_map(int fd, off_t offset, size_t size, int expand,
...
+ * UIO map offsets are offsets inside a page. The Linux UIO mmap path maps
+ * from the page boundary and exposes the usable region at raw mapping plus
+ * this offset.
+ *
+ * @param[in]	dev_name UIO-backed device name.
+ * @param[in]	index UIO map index.
+ * @param[in]	offset UIO map offset.
+ * @return	0 on success, or -errno on error.
+ */
+int metal_linux_uio_validate_offset(const char *dev_name,
```
arnopo: do you really need to expose this API to the application?

lib/system/linux/irq.c
```diff
@@ -266,10 +266,48 @@ void metal_linux_irq_shutdown(void)
...
+	}
+
+	metal_mutex_acquire(&irq_lock);
+	irqs_devs[irq] = NULL;
+	metal_bitmap_clear_bit(irqs_enabled, irq - linux_irq_cntr.irq_base);
+	irqs[irq].hd = NULL;
+	irqs[irq].arg = NULL;
+	metal_mutex_release(&irq_lock);
+
+	metal_linux_irq_notify();
```
arnopo: code above  should not be done here you should only have a check

if (metal_linux_irq_is_enabled(irq)
    return EINVAL;

=> the standard `metal_irq_disable` should be called before by the IRQ consumer

lib/system/linux/device.c
```diff
@@ -77,6 +83,55 @@ static struct linux_device *to_linux_device(struct metal_device *device)
 	return metal_container_of(device, struct linux_device, device);
 }
 
+static bool metal_linux_is_uio_bus(const struct linux_bus *lbus)
+{
+	return strcmp(lbus->bus_name, "uio") == 0;
+}
+
+static int metal_linux_copy_string(char *output, size_t output_len,
```
arnopo: why  creating this function and not directly using `snprintf` or `memcpy`?

lib/system/linux/device.c
```diff
@@ -77,6 +83,55 @@ static struct linux_device *to_linux_device(struct metal_device *device)
...
+				   const char *input)
+{
+	int result;
+
+	if (!output || !output_len || !input)
+		return -EINVAL;
+
+	result = snprintf(output, output_len, "%s", input);
+	if (result < 0 || result >= (int)output_len)
```
arnopo: if result >= (int)output_len) there is probably some memories that are corrupted

lib/system/linux/device.c
```diff
@@ -168,6 +243,80 @@ int metal_linux_uio_map_info(const char *dev_name,
...
+	char value[PATH_MAX];
+	bool found = false;
+	int result = -ENODEV;
+
+	if (!uio_name || !strlen(uio_name) || !ldev)
+		return -EINVAL;
+
+	dir = opendir(METAL_UIO_CLASS_PATH);
+	if (!dir)
+		return -errno;
```
arnopo: seems not standard error

lib/system/linux/device.c
```diff
@@ -168,6 +243,80 @@ int metal_linux_uio_map_info(const char *dev_name,
...
+						 sizeof(ldev->uio_name),
+						 value);
+		if (result)
+			goto out;
+		result = metal_linux_copy_string(ldev->uio_dev_name,
+						 sizeof(ldev->uio_dev_name),
+						 entry->d_name);
+		if (result)
+			goto out;
+	}
```
arnopo, line 409: Adding few comments in code would help to understand this function, you can also add an header that explains it

lib/system/linux/sys.h
arnopo: why functions are defineded in device.c and declared in sys.h?

lib/system/linux/device.c
```diff
@@ -100,6 +102,72 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
 	return 0;
 }
 
+int metal_linux_uio_validate_offset(const char *dev_name,
+				    unsigned int index,
```
arnopo: index seems useless as only use to print the error message,

lib/system/linux/device.c
```diff
@@ -450,6 +450,8 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 		/* Try and open the device. */
 		error = ldrv->dev_open(lbus, ldev);
 		if (error) {
+			if (open_error == -ENODEV || error != -ENODEV)
+				open_error = error;
```
arnopo: seems to me that it is not the good strategy .

1) if ldrv->dev_open return an error, the device is probably not opened so no need to close it
2) if the open fails, we should keep the error and close all devices aready opened

```
for_each_linux_driver(lbus, ldrv) {

		/* Check if we have a viable driver. */
		if (!ldrv->sdrv || !ldrv->dev_open)
			continue;

		/* Reset device data. */
		memset(ldev, 0, sizeof(*ldev));
		strncpy(ldev->dev_name, dev_name, sizeof(ldev->dev_name) - 1);
		ldev->fd = -1;
		ldev->ldrv = ldrv;
		ldev->device.bus = bus;

		/* Try and open the device. */
		error = ldrv->dev_open(lbus, ldev);
		if (error) {
			goto close_dev;
		}

		*device = &ldev->device;
		(*device)->name = ldev->dev_name;

		metal_list_add_tail(&bus->devices, &(*device)->node);
		return 0;
	}

close_dev:
    for_each_linux_driver(lbus, ldrv) {
		ldev->ldrv = ldrv;
		ldrv->dev_close(lbus, ldev);
        metal_list_del(&ldev->device.node);
	}
	free(ldev);

	return error;
```

arnopo (commented):
@bentheredonethat 
Sorry for the delay, please find some comment.
I need to past time on "lib: linux: add UIO bus open by class name" to understand your work. adding more comment would help me

lib/system/linux/sys.h
bentheredonethat: metal_linux_uio_validate_offset() will be static in lib/system/linux/device.c

lib/system/linux/irq.c
bentheredonethat: ok will fix. 

metal_linux_irq_unregister_dev() will only detach the Linux fd-to-device bookkeeping after the consumer has disabled the IRQ.

lib/system/linux/device.c
bentheredonethat: ok will remove and replace with direct snprintf() with bounds checking

lib/system/linux/device.c
bentheredonethat: ok deleting the helper containing that exact code

lib/system/linux/device.c
bentheredonethat: Addressed by mapping missing /sys/class/uio to -ENODEV, while preserving other opendir() failures as -errno. This keeps “UIO class unavailable” consistent with the rest of Linux bus probing.

lib/system/linux/device.c
bentheredonethat: ok adding comments

lib/system/linux/sys.h
bentheredonethat: will fix as follows:

no longer a cross-file/internal API. The function is local to the file that uses it, which resolves the mismatch between declaration location and implementation location

lib/system/linux/device.c
bentheredonethat: will remove

lib/system/linux/device.c
bentheredonethat: ok will do this - thanks

lib/system/linux/device.c
```diff
@@ -151,17 +405,119 @@ static int metal_uio_dev_bind(struct linux_device *ldev,
...
 	struct metal_io_region *io;
-	struct dlist *dlist;
+	struct metal_uio_map_info map_info;
+	size_t map_len, region_size;
 	int result, i;
-	void *virt;
+	void *raw, *virt;
 	int irq_info;
 
+	i = 0;
```
arnopo: can be initialized when  declared

lib/system/linux/device.c
```diff
@@ -151,17 +405,119 @@ static int metal_uio_dev_bind(struct linux_device *ldev,
...
+	ldev->fd = result;
+
+	metal_log(METAL_LOG_DEBUG, "opened %s:%s as %s\n",
+		  lbus->bus_name, ldev->dev_name, ldev->dev_path);
+
+	for (i = 0; i < METAL_MAX_DEVICE_REGIONS; i++) {
+		phys = &ldev->region_phys[ldev->device.num_regions];
+		result = metal_uio_read_map_attr(ldev, i, "offset", &offset);
+		if (result)
+			break;
```
arnopo, line 523: Why do you break for this one?

lib/system/linux/device.c
```diff
@@ -151,17 +405,119 @@ static int metal_uio_dev_bind(struct linux_device *ldev,
...
+		i++;
+	} while (i < 1000);
+	if (i >= 1000) {
+		metal_log(METAL_LOG_ERROR, "failed to open file %s, timeout.\n",
+			  ldev->dev_path);
+		return -ENODEV;
+	}
+	result = metal_open(ldev->dev_path, 0);
+	if (result < 0) {
+		metal_log(METAL_LOG_ERROR, "failed to open device %s\n",
```
arnopo:
```suggestion
		metal_log(METAL_LOG_ERROR, "failed to open device %s: %s\n",
```

lib/system/linux/device.c
```diff
@@ -736,6 +969,16 @@ static int metal_linux_probe_bus(struct linux_bus *lbus)
 	struct linux_driver *ldrv;
 	int ret, error = -ENODEV;
 
+	/*
+	 * /sys/class/uio is a class, not a bus. Register the synthetic bus only
+	 * when the UIO class exists and skip normal bus/driver probing.
+	 */
```
arnopo: Comment not clear to me .  the sentence "/sys/class/uio is a class, not a bus" is confusing. 
suggestion 
```
	/*
	 * Register the synthetic bus only  when the  /sys/class/uio 
	 *class exists and skip normal bus/driver probing.
	 */
```

lib/system/linux/device.c
```diff
@@ -231,21 +326,52 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
 	metal_log(METAL_LOG_DEBUG, "opened %s:%s as %s\n",
 		  lbus->bus_name, ldev->dev_name, ldev->dev_path);
 
-	for (i = 0, result = 0; !result && i < METAL_MAX_DEVICE_REGIONS; i++) {
+	for (i = 0; i < METAL_MAX_DEVICE_REGIONS; i++) {
 		phys = &ldev->region_phys[ldev->device.num_regions];
+		result = metal_uio_read_map_attr(ldev, i, "offset", &offset);
+		if (result)
+			break;
```
arnopo, line 523: Error case to manage to properly leave this function here and below , closing /unregistering, freeing, ... thinks

lib/system/linux/device.c
```diff
@@ -92,6 +98,48 @@ static struct linux_device *to_linux_device(struct metal_device *device)
...
+	}
+
+	fclose(fp);
+
+	newline = strchr(output, '\n');
+	if (newline)
+		*newline = '\0';
+
+	return 0;
+}
```
arnopo, line 167: suggestion ( can also be applyed for  `metal_uio_read_str_attr()`)
```
static int metal_linux_read_first_line(const char *path, char *output,
				       size_t output_len)
{
	FILE *fp;
	char *newline;
    err = 0;

	if (!path || !output || output_len < 2)
		return -EINVAL;

	fp = fopen(path, "r");
	if (!fp)
		return -errno;

	if (!fgets(output, output_len, fp)) {
		int err = ferror(fp) ? -errno : -ENODATA;

		goto close_file;
	}

	newline = strchr(output, '\n');
	if (newline)
		*newline = '\0';

close_file:
	fclose(fp);

	return err;
}
```

lib/system/linux/device.c
```diff
@@ -192,6 +264,96 @@ static int metal_linux_uio_map_info(struct metal_uio_map_info *info)
...
+			continue;
+
+		if (found) {
+			/* Duplicate names cannot be opened deterministically. */
+			result = -EEXIST;
+			goto out;
+		}
+		found = true;
+
+		result = snprintf(ldev->cls_path, sizeof(ldev->cls_path),
```
arnopo, line 379: not obvious what is cls_path , pleas clarify by comments or by describing ldev fields

lib/system/linux/device.c
```diff
@@ -390,6 +507,98 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
...
+		metal_log(METAL_LOG_ERROR, "device %s:%s not found\n",
+			  lbus->bus_name, ldev->dev_name);
+		return -ENODEV;
+	}
+	metal_log(METAL_LOG_DEBUG, "opened sysfs device %s:%s\n",
+		  lbus->bus_name, ldev->dev_name);
+
+	result = metal_uio_dev_bind(ldev, ldrv);
+	if (result)
+		return result;
```
arnopo: error case to manage to free resources here and below

arnopo (changes requested):
Some of  the commits are still not easy to review. Reverse engineering is required to understand it.
Adding more details on the algorithm you try to apply in the commit message could help

lib/system/linux/device.c
bentheredonethat: will fix

lib/system/linux/device.c
bentheredonethat: will fix

lib/system/linux/device.c
bentheredonethat: will fix

lib/system/linux/device.c
bentheredonethat: will fix

lib/system/linux/device.c
bentheredonethat: will fix

lib/system/linux/device.c
bentheredonethat: will add comment

lib/system/linux/device.c
bentheredonethat: will fix

lib/system/linux/device.c
bentheredonethat: ok will fix

bentheredonethat:
@arnopo  i split further the UIO rework so there are more comments, split the commits for easier review, and detail the algorithm used as well as the nits

lib/system/linux/device.c
```diff
@@ -461,9 +464,15 @@ static int metal_linux_dev_open(struct metal_bus *bus,
 		return 0;
 	}
 
+	goto out;
+
+close_dev:
+	if (ldrv->dev_close)
```
arnopo: you should have for_each_linux_driver(lbus, ldrv) here , no?

lib/system/linux/device.c
```diff
@@ -100,6 +117,81 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
...
+	int result;
+
+	if (!info || !info->phys || !info->map_len || !info->region_size)
+		return -EINVAL;
+
+	result = metal_linux_uio_validate_offset(info->dev_name, info->offset);
+	if (result)
+		return result;
+
+	if (!info->map_size || info->offset >= info->map_size) {
```
arnopo: if `info->map_size == 0`  the condition info->offset >= info->map_size is also true
```suggestion
	if (info->offset >= info->map_size) {
```

lib/system/linux/device.c
```diff
@@ -100,6 +117,81 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
...
+		metal_log(METAL_LOG_ERROR,
+			  "device %s has invalid UIO size 0x%lx for offset 0x%lx\n",
+			  info->dev_name ? info->dev_name : "<unknown>",
+			  info->map_size, info->offset);
+		return -EINVAL;
+	}
+
+	if ((unsigned long)(size_t)info->map_size != info->map_size) {
```
arnopo:
```suggestion
	if (info->map_size >sizeof(size_t) {
```

lib/system/linux/device.c
```diff
@@ -100,6 +117,81 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
...
+		metal_log(METAL_LOG_ERROR,
+			  "device %s UIO size 0x%lx overflows size_t\n",
+			  info->dev_name ? info->dev_name : "<unknown>",
+			  info->map_size);
+		return -EOVERFLOW;
+	}
+
+	if (info->map_addr > (metal_phys_addr_t)-1 - info->offset) {
```
arnopo:
```suggestion
	if (info->map_addr + info->offset < info->map_addr) 
```

lib/system/linux/device.c
arnopo, line 523: some fixes for the error management should be present in this commit  not in the one where you create 
`metal_uio_populate`

lib/system/linux/device.c
```diff
@@ -411,7 +411,14 @@ static void metal_uio_dev_close(struct linux_bus *lbus,
 		ldev->sdev = NULL;
 	}
 	if (ldev->fd >= 0) {
+		/*
+		 * Disable first so unregister only removes device bookkeeping;
+		 * IRQ handler teardown remains in the generic IRQ path.
+		 */
+		metal_irq_disable(ldev->fd);
+		metal_linux_irq_unregister_dev(ldev->fd);
```
arnopo: should the iRQ disable first in this function befor unmapping the memory?

lib/system/linux/device.c
```diff
@@ -388,6 +382,127 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
...
+	 * calls dev_close() to release parent sysfs and driver override state.
+	 */
+
+	/*
+	 * Parent-bus opens still need the requested platform or PCI device
+	 * bound to the selected UIO driver before a /dev/uioX node can exist.
+	 */
+	result = metal_uio_dev_bind(ldev, ldrv);
+	if (result)
+		return result;
```
arnopo: you should close the sysfs device here and after on error

lib/system/linux/device.c
```diff
@@ -388,6 +382,127 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
...
+
+	/*
+	 * A bound parent device exposes one UIO child below its sysfs device
+	 * directory. Use that child name to derive both sysfs and /dev paths.
+	 */
+	result = snprintf(path, sizeof(path), "%s/uio", ldev->sdev->path);
```
arnopo, line 651: snprintf return an negative value on error  that should be tested

bentheredonethat:
Thanks @arnopo, I addressed the review comments across the series.

  Summary of updates:

  - In `lib: linux: preserve device-open errors`, `metal_linux_dev_open()` now closes a failed backend driver and continues trying remaining viable drivers, while preserving the first useful errno
  to return if all drivers fail.

  - In `lib: linux: fix UIO mmap offset handling`:
    - Removed the redundant `!info->map_size` check since `info->offset >= info->map_size` already catches `map_size == 0`.
    - Replaced the `size_t` overflow check with an explicit `SIZE_MAX` comparison and added the needed `<stdint.h>` include.
    - Switched the physical-address overflow check to the suggested wraparound-style check.
    - Moved the UIO map-read error handling into this commit: missing next map ends the loop, while other map-read errors fail open.

  - In `lib: linux: clear UIO IRQ bookkeeping on close`, IRQ disable/unregister now happens before unmapping device regions, and close resets the region count after unmapping.

  - In `lib: linux: factor common UIO populate path`:
    - Parent UIO open error paths after `sysfs_open_device()` now clean up locally.
    - Cleanup remains idempotent so the common open path can safely call `dev_close()` again.
    - Added negative-return checks for the touched `snprintf()` calls.

lib/system/linux/device.c
bentheredonethat: current latest code:
  - result < 0 handles snprintf() failure.
  - result >= (int)sizeof(path) handles truncation.
  - goto fail routes through cleanup, including closing the opened sysfs device via metal_uio_dev_close(lbus, ldev).

lib/system/linux/device.c
bentheredonethat: Addressed -  if metal_uio_dev_bind() fails, the code no longer returns directly. It jumps to fail, which closes the sysfs device. The same cleanup is used for later errors after bind as well.

lib/system/linux/device.c
bentheredonethat: Added comment

Current metal_uio_dev_close() does IRQ disable/unregister before the region unmap loop:

`
  if (ldev->fd >= 0) {
  	/*
  	 * Disable first so unregister only removes device bookkeeping;
  	 * IRQ handler teardown remains in the generic IRQ path.
  	 */
`

lib/system/linux/device.c
bentheredonethat: behavior is no longer “any offset read error means end of maps.” It is now:

  - -ENOENT: normal end of UIO map list.
  - anything else: real open failure, go through cleanup.

lib/system/linux/device.c
bentheredonethat: applied

lib/system/linux/device.c
bentheredonethat: Went with check of 

>   if (info->map_size > SIZE_MAX) {
>   	metal_log(METAL_LOG_ERROR,
>   		  "device %s UIO size 0x%lx overflows size_t\n",
>   		  info->dev_name ? info->dev_name : "<unknown>",
>   		  info->map_size);
>   	return -EOVERFLOW;
>   }

sizeof(size_t) is the number of bytes in the type, usually 4 or 8, not the largest value a size_t can represent. SIZE_MAX is the right constant for the intended  overflow check.

lib/system/linux/device.c
bentheredonethat: applied

lib/system/linux/device.c
bentheredonethat: NA.

 Current metal_linux_dev_open() has a single for_each_linux_driver(lbus, ldrv) loop, and failed opens clean up the same driver that just attempted the open.

There is no post-loop close_dev: block anymore, so there is no need to add another for_each_linux_driver() there.

lib/system/linux/device.c
```diff
@@ -100,6 +119,81 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
...
+	}
+
+	return 0;
+}
+
+/*
+ * Translate UIO sysfs map attributes into the values libmetal needs:
+ * the mmap() length for cleanup, the usable physical start address, and
+ * the usable I/O region size after skipping the map offset.
+ */
```
arnopo:
```suggestion
/**
 * @internal
 * 
 * @brief Translate UIO sysfs map attributes into libmetal map information.
 *
 * This fills in the values required by libmetal:
 * - the mmap() length used for cleanup,
 * - the usable physical start address,
 * - the usable I/O region size after skipping the map offset.
 *
 * @param info  Pointer to the map information structure to populate.
 *
 * @return 0 on success, or a negative error code on failure.
 */
```

lib/system/linux/device.c
```diff
@@ -100,6 +119,81 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
 	return 0;
 }
 
+/*
+ * Validate the sysfs map offset before it is applied to the mmap() base.
+ * The Linux UIO ABI exposes one mmap slot per page-sized index, so the
+ * per-map offset must stay inside a single host page.
+ */
```
arnopo, line 242:
```suggestion
/**
 * @internal
 * 
 * @brief Validate the sysfs map offset before it is applied to the mmap() base.
 *
 * The Linux UIO ABI exposes one mmap slot per page-sized index, so the
 * per-map offset must remain within a single host page.
 *
 * The offset is applied inside one page returned by mmap().
 * Larger offsets cannot be represented by adjusting the returned mapping.
 *
 * @param dev_name  Device name used for error reporting; may be NULL.
 * @param offset    Offset to validate, in bytes.
 *
 * @return 0 on success, or -EINVAL if the offset exceeds the host page size.
 */
```

lib/system/linux/irq.c
```diff
@@ -266,10 +266,69 @@ void metal_linux_irq_shutdown(void)
...
 		return;
 	}
 	irqs_devs[irq] = dev;
 }
+
+/*
+ * Drop the device pointer associated with a Linux IRQ fd during device close.
+ * The caller must disable the IRQ first so the dispatch path cannot observe
+ * an enabled IRQ whose owning device has already been detached.
+ */
```
arnopo: Redundant comment here and for other functions below . doxygen documentation in irq.h already documents the functions

lib/system/linux/device.c
```diff
@@ -119,6 +125,32 @@ static int metal_uio_read_map_attr(struct linux_device *ldev,
 	return 0;
 }
 
+/*
+ * Read string-valued UIO sysfs attributes such as /sys/class/uio/uioX/name.
+ * The value is explicitly terminated so callers can compare it as a C string.
+ */
```
arnopo, line 205: Use doxygen format with  `@internal` tag for the functions you introduced in this commit

lib/system/linux/device.c
```diff
@@ -438,6 +403,125 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
...
+			goto close_list;
+		}
+		break;
+	}
+	result = 0;
+
+close_list:
+	sysfs_close_list(dlist);
+	if (result)
+		goto fail;
```
arnopo: Not really readable. Avoid such kind of label and use them only for error management.
here remove the label and the test on result, just keeping `sysfs_close_list(dlist);`

the at the end of the function:

```c
close_list:
	sysfs_close_list(dlist);

fail:
	metal_uio_dev_close(lbus, ldev);
	return result;

```

lib/system/linux/device.c
arnopo: There is an imbalance between the open and close operations in the function because for_each_linux_driver(lbus, ldrv) handles the open part. Therefore, either remove for_each_linux_driver or close all opened devices if an error occurs.

lib/system/linux/irq.c
```diff
@@ -266,10 +266,69 @@ void metal_linux_irq_shutdown(void)
...
+	metal_mutex_acquire(&irq_lock);
+	/*
+	 * Unregister only detaches the device association. The IRQ handler and
+	 * enabled state remain owned by metal_irq_disable()/unregister().
+	 */
+	if (metal_bitmap_is_bit_set(irqs_enabled, offset)) {
+		metal_mutex_release(&irq_lock);
+		return -EINVAL;
+	}
+	irqs_devs[irq] = NULL;
```
arnopo, line 300: seems that you should add protection on this variable in `metal_linux_irq_register_dev` and `metal_linux_irq_handling` and `metal_linux_irq_handling`

bentheredonethat:
@arnopo 

addressed as follows:

  - For the UIO mmap offset helpers, I converted the comments for `metal_linux_uio_validate_offset()` and `metal_linux_uio_map_info()` to Doxygen format with `@internal`, `@brief`, parameter
  documentation, and return value documentation. I also documented `struct metal_uio_map_info` in the same style since it is part of that helper flow.

  - For the IRQ bookkeeping helpers, I removed the redundant function comments from `lib/system/linux/irq.c` because the functions are already documented in `lib/system/linux/irq.h`. I kept only the
  local inline comment that explains the teardown ordering/ownership inside the implementation.

  - For the UIO helper functions introduced in the common populate/class-name/synthetic-bus commits, I converted the function-level comments to Doxygen with `@internal`. This covers the newly
  introduced helpers in `device.c`, including the string/sysfs helpers, class-name lookup, common populate path, synthetic UIO bus helper, and synthetic-bus probe/close behavior.

  - For the `close_list:` readability issue, I reworked the flow so the normal path closes the sysfs list directly and continues, while `close_list:` is used only as an error cleanup label before
  the shared failure path. This removes the label-plus-result-test pattern and keeps labels focused on error management.

  - For the `metal_linux_dev_open()` open/close imbalance concern, I changed the retry logic so only clean `-ENODEV` backend misses continue to later drivers. If a backend finds the device but fails
  with a real error, the attempted backend is cleaned up and that errno is returned instead of continuing through the driver list.

  - I also removed duplicate inline explanation where the new Doxygen comments now carry the same information, especially around the synthetic UIO bus probe/close behavior.

lib/system/linux/irq.c
bentheredonethat: fixed in latest

arnopo (approved):
Look good to go

arnopo:
@edmooring @tnmysh , could you have a look please

nathalie-ckc:
2026-07-15 System Reference call:
@tnmysh will take a look at this in about a week's time.

lib/system/linux/device.c
```diff
@@ -218,34 +347,72 @@ static int metal_uio_dev_open(struct linux_bus *lbus, struct linux_device *ldev)
...
-		return result;
+		goto fail;
 	}
 	ldev->fd = result;
 
 	metal_log(METAL_LOG_DEBUG, "opened %s:%s as %s\n",
 		  lbus->bus_name, ldev->dev_name, ldev->dev_path);
 
-	for (i = 0, result = 0; !result && i < METAL_MAX_DEVICE_REGIONS; i++) {
+	for (i = 0; i < METAL_MAX_DEVICE_REGIONS; i++) {
```
tnmysh, line 364: This syntax error should be fixed where it was introduced.

tnmysh: oh my bad! It's not syntax error. Please ignore this comment.

tnmysh (approved):
Looks good to go. This is much better now than before.

Thank You.

PR #370, opened 2026-09-04
linux: remove sudo escalation from probe_driver and fix CI test setup

cshilwant:
1. Removes the `sudo modprobe` fallback from `metal_linux_probe_driver()`                                                                                                                                                                    
    in `lib/system/linux/device.c`. A library must not silently attempt                                                        
    privilege escalation on behalf of its caller - this is a security                                                                                                                                                                         
    concern and fails on embedded/production systems where `sudo`
    is absent or not configured for password-less use.

2. Updates the CI nonreg test workflow to explicitly load the required                                                                                                                                                                       
    UIO kernel modules (`uio`, `uio_pdrv_genirq`, `uio_dmem_genirq`)                                                                                                                                                                          
    before running `make test`, since the environment (& not the library)                                                                                                                                                                    
    is responsible for module loading

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -68,6 +68,7 @@ jobs:
         make
     - name: execute test
       run: |
+        sudo modprobe uio uio_pdrv_genirq uio_dmem_genirq
         cd build
         make test
     - name: test logs
--- a/lib/system/linux/device.c
+++ b/lib/system/linux/device.c
@@ -1113,21 +1113,6 @@ static int metal_linux_probe_driver(struct linux_bus *lbus,
 		ldrv->sdrv = sysfs_open_driver(lbus->bus_name, ldrv->drv_name);
 	}
 
-	/* Try sudo probing the module and then open the driver. */
-	if (!ldrv->sdrv) {
-		ret = snprintf(command, sizeof(command),
-			       "sudo modprobe %s > /dev/null 2>&1", ldrv->mod_name);
-		if (ret >= (int)sizeof(command))
-			return -EOVERFLOW;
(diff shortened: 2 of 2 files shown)


glneo (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -68,6 +68,7 @@ jobs:
         make
     - name: execute test
       run: |
+        sudo modprobe uio uio_pdrv_genirq uio_dmem_genirq
         cd build
```
arnopo, line 72: Could you take opportunity to remove the `cd build `that is already done during `build` step ?

arnopo (approved):
Tested with success on my PC, 
Just a comment to request an extra fix of an issue i found during my test. Not mandatory, I can address it later

bentheredonethat (approved):
LGTM

.github/workflows/continuous-integration.yml
cshilwant: > Each run keyword represents a new process and shell in the runner environment

@arnopo  Thanks for the suggestion! As per the GitHub Actions workflow syntax reference (`jobs.<job_id>.steps[*].run` - https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun), doesn't each `run:` step spawn a new shell process, meaning the `cd build` from the build step would not carry over to the execute test step?

Probably also the reason why the test logs step references the full path `build/Testing/Temporary/LastTest.log` - https://github.com/OpenAMP/libmetal/blob/main/.github/workflows/continuous-integration.yml#L76 rather than just `Testing/Temporary/LastTest.log` ?

